PR #8, opened 2018-01-17
Add contract api for design_a

mossid:


spec/design_a.md
```diff
@@ -48,3 +48,48 @@ reflect that someone wants to send tokens from Ethereum. Every subsequent
 node adds another confirmation to the peg zone state. Every BeginBlock
 invocation the peg zone checks whether any incoming Ethereum transfers have 
 reached a super-majority of confirmations and if so creates an IBC packet.
+
+## API
+
+### Ethereum Smart Contract
+
+#### update(address[] newAddress, uint64[] newPower, /*bytes32 nonce,*/ uint8[] v, bytes32[] r, bytes32[] s)
```
zmanian: So the nonce wouldn't be passed in as argument. it would be internal state of the function. It would need to be signed.

I suspect with need a `getNonce` function here. I also think nonces should be `uint32` vs bytes32

design-a/contracts/Valset.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    function verify(bytes32 hash, uint8[] v, bytes32[] r, bytes32[] s) returns (bool) {
+        if (!(v.length == r.length &&
+              r.length == s.length)) {
+            return false;
+        }
+
+        uint64 signedPower = 0;
+
+        for (uint i = 0; i < v.length; i++) {
```
zmanian: Here is what i don't like about this algorithm, let's says that signatures for val 1, 3, 5 are enough voting power to be > 2/3rd's of the voting power.

I still need to submit signatures for 2 and 4 too.

Consider submitting the indexes of validators in the arguments so we only need to submit the minimum number of signatures.

PR #14, opened 2018-01-25
Start pegzone

ethanfrey:
Ripped out some pieces from cosmos-sdk v0.8 to use as a basis.
I know it is a bit old, but it is stable, and should be enough to make a prototype.


adrianbrink:
I'm closing this for now. Let's try to keep lingering PRs to a minimum.

PR #27, opened 2018-02-13
Ethereum SC  #1

fedekunze:
Hi, 

I'm having a InvalidSignature issue on the `test/valset.js` file. I implemented the response from this [thread](https://ethereum.stackexchange.com/questions/12621/getting-the-wrong-address-back-from-ecrecover/12684) and works but the idea is to test it with the contract, which I haven't yet archived.

According to this [source](https://ethereum.stackexchange.com/questions/15364/ecrecover-from-geth-and-web3-eth-sign?noredirect=1&lq=1) I think I'm handling the signature correctly on smart contract, but not on the test files. I've tried using [ethereumjs-util](https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/index.md) and `web3` with no results at the moment.


sunnya97:
closed in favour of #31

PR #28, opened 2018-02-17
PeggyApp Implementation

adrianbrink:
This is the application that is written on top of the cosmos-sdk. It is the first major piece of software that is written from scratch just on the cosmos-sdk and as such provides an ideal testing ground for the ergonomics of the sdk. Everything we learn here, we need to feed back into the cosmos-sdk in order to improve it.

The overall design is simple. All transactions are defined in types/transactions.go. 
The three main components are `WitnessTxMapper`, `WithdrawTxMapper` and the `bank module`.

`WitnessTxMapper` stores a mapping from `WitnessTxHash -> WitnessData`. The first `WitnessTx` initializes that mapping and subsequent `WitnessTx` transactions will append information to the respective `WitnessData`. Every `EndBlock` we check whether >2/3 of validators have witnessed a `Lock Event` and if so, we credit that amount.

`WithdrawTxMapper` stores a mapping from `WithdrawTxHash -> WithdrawData`. It works like the `WitnessTxMapper` except that it is used to accumlate `SignTx` transactions so that a `Relayer` can eventually take them and submit them to `Peggy SC`.

The `bank module` keeps track of user balances and adds the ability to send tokens between accounts.

At the moment the `Withdraw` and `Witness` are just placed in the global state. There might be a point to have them as separate modules and move them into `x/withdraw` and `x/witness`. 

Currently, this PR does not compile. It adds the overall skeleton and architecture and now we have to go through it and fix all the broken parts. Ideally, we should start with the types and work our way up. In order to do this, we have to figure out how to best test modules in isolation and generally how to test applications without having to run against a live instance of TMC.

Let's use this opportunity to test how easy it is to build on top of the cosmos-sdk and how we can smooth out the rough edges.

This PR is not designed to build any of the other processes, such as `Signer`, `Witness`, or `Relayer`. Those will be added in different PRs.

@mossid @sunnya97 @zmanian : Once #27 is merged, let's focus on building the peg zone. #27 includes all the needed functionality for the MVP on `Ethereum SC`. After the peg zone application is merged, we can move on to the `Witness`, then to the `Signer` and finally, we can implement the `Relayer`. If we see that it is infeasible to work with 4 people on the same code, then we can parallelise the development and build the `Witness` and the `PeggyApp` at the same time.

Please let me know if you have suggestions for this development plan. Any two people can merge a PR. So, you need one other person (from us 4) to review and approve the PR and you are free to merge it. This applies to every part of `cosmos/peggy`.


mossid:
```
message ResponseEndBlock {
  repeated Validator validator_updates = 1 [(gogoproto.nullable)=false];
  ConsensusParams consensus_param_updates = 2;
}
```

The return type of `EndBlocker` is `abci.ResponseEndBlock` and it does not contain any information about the storage. I think we should check +2/3 in the handler.

tx.proto
```diff
@@ -0,0 +1,35 @@
...
+/*
+message Tx {
+    oneof tx {
+        WitnessTx witness = 1;
+        SignTx    sign    = 2;
+        RelayTx   relay   = 3;
+    }
+}
+
+message RelayTx {
```
adrianbrink: We don't need a RelayTx as far as I can tell. 

The relayer will just passively read the latest state from the peg zone by listening to events and then send it to Ethereum.

tx.proto
```diff
@@ -0,0 +1,35 @@
...
+
+message WitnessTx {
+    bytes signature = 1;
+    uint64 sequence = 2;
+    oneof tx {
+        LockMsg lock = 3;
+    }
+}
+
+message LockMsg {
```
adrianbrink: What's the difference between a LockMsg and a WitnessTx?

mossid: WitnessTx was intended to be a wrapper for LockMsg and BurnMsg. Since we are working with LockMsg for now, WitnessTx will only contain LockMsg.

Also, WitnessTx corresponds to sdk.Tx, LockMsg corresponds to sdk.Msg.

PR #31, opened 2018-02-17
All preliminary contracts written and tested

sunnya97:


adrianbrink:
This looks great. Let's go through this when you are in the office.

I'd like some clarity on the setup for the migrations, deploying this to a network and running the tests. Then let's update the readme and then we can merge this.

PR #37, opened 2018-03-02
Feature/peg zone witness

mossid:
There is a modification of the design of #28.

To separate the signer's address from the information specific to the msg type, WitnessMsg had become a union of crypto.Address and WitnessInfo. It made possible to easily implement GetSignBytes. The binary form of an Info is used for the mapper's key.

app_test.go is working test file.


fedekunze:
outdated

Issue #35, opened 2018-02-21
Event logs should contain only static types


mossid:
When dynamic types like `bytes` are included in a log, it makes the length of the log `data` not a multiple of 32. For example, this is a log data for `Lock(bytes to, uint64 amount, address token)`(line break for every 32 bytes == 1 word).

```
0000000000000000000000000000000000000000000000000000000000000060
0000000000000000000000000000000000000000000000000000000000002710
0000000000000000000000000000000000000000000000000000000000000000
0000000000000000000000000000000000000000000000000000000000000003
686868
```

The first word contains the relative position of the first element(hex 60 == dec 96). The word in that position contains the length of the `bytes`. The 3-byte data `686868` follows it.

It makes the `parse_log` in ethabi package not works. (ethabi/src/util.rs line 7-9)

```
if data.len() % 32 != 0 {
    return Err(ErrorKind::InvalidData.into());
}
```

To solve this problem, we have to change every `bytes` in solidity files into `bytes32`. On the frontend side, we can either 1. hash the input data to make it bytes32 or 2. accept bytes only whose length is less than 32.

So the api should be modified as:

```
lock(bytes32 to, uint64 value, address token) external payable returns (bool);
event Lock(bytes32 to, uint64 value, address token);

unlock(address[2] addressArg, uint64 value, bytes32 chain, uint16[] idxs, uint8[] v, bytes32[] r, bytes32[] s) external returns (bool)
```

`unlock` also has to take bytes32 in favor of #32

PR #38, opened 2018-03-05
move bytes to bytes32, add nonce in Lock()

mossid:
To read event logs from Witness app, events should contain only static fields. See #35. The contract, spec, and test are modified.

To distinguish events that have same fields, a nonce is added to Lock() event. Internal `nonce` variable is increased each time lock() is called.


fedekunze:
outdated

PR #50, opened 2019-04-30
Unidirectional ethereum-contracts

denalimarsh:
Changes to Peggy's ethereum contracts. Contracts only include functionality for unidirectional transfers of funds from Ethereum -> Cosmos and the branch is being actively used for testing transfer of assets from Ethereum to Cosmos. Existing work on secured unlocking in ```Valset.sol``` will be reincorporated once one-way transfers have been prototyped. If these changes are too breaking for master, wait and merge this PR once existing features have been reintegrated.


cwgoes:
Thanks! Adding a written technical specification in Markdown would be helpful for review here.

denalimarsh:
> Thanks! Adding a written technical specification in Markdown would be helpful for review here.

Added a project specification in the README.md of the `/ethereum-contracts` directory.

fedekunze:
cc: @okwme can you review as well ?

ethereum-contracts/contracts/Peggy.sol
```diff
@@ -1,95 +1,266 @@
...
+        onlySender(_id)
+        canDeliver(_id)
+        external
+        returns (bool)
+    {
+        require(isLocked(_id));
 
-        bytes32 hashData = keccak256(name, decimals);
-        require(Valset.verifyValidators(hashData, signers, v, r, s));
+        // Transfer item's funds and unlock it
```
fedekunze: what's an item in this context ?

ethereum-contracts/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "ethereum-contracts",
+  "name": "ethreum-contracts",
```
fedekunze: typo ?
```suggestion
  "name": "ethereum-contracts",
```

denalimarsh: Fixed, thanks for catching that.

ethereum-contracts/contracts/Peggy.sol
denalimarsh: An item refers to a single incoming transaction containing funds to be locked. Originally named 'escrow', the terminology was updated to 'item' to prevent confusion regarding the fact that the funds are not technically held in escrow for the unidirectional implementation. Happy to update the naming conventions again, should there be consensus about a more suitable name for an instance of locked funds.

fedekunze: I see. We use the term `Msg` on the SDK (See [`sdk.Msg`](https://github.com/cosmos/cosmos-sdk/blob/master/types/tx_msg.go) and [`auth.StdTx`](https://github.com/cosmos/cosmos-sdk/blob/master/x/auth/stdtx.go#L22) for reference). I think that's a more accurate term for this case

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
...
-  - Ganache: Download file [here](http://truffleframework.com/ganache/)
+## Project Background
+We are hoping to create a closed system for intra network transfers of cryptocurrency between blockchains, spearheaded by a proof-of-concept which enables secured transactions between Ethereum and Cosmos.
 
-## Instructions
+## Smart Contract Scope
 
-You can test the contracts using Truffle Testrpc in the console or Ganache UI.
+### Goals of the Smart Contracts
+1. Securely implement core functionality of the system such as asset locking and event emission without endangering any user funds. As such, this prototype does not permanently lock value and allows the original sender full access to their funds at any time.
```
cwgoes: What do you mean by "permanently lock"? Can the original sender unlock their original tokens even after they've sent tokens to Cosmos?

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
...
+3. Fully gas optimize and streamline operational functionality; ease and clarity of testing has been favored over some gas management and architectural best practices.
 
+## Ethereum Cosmos Bridge Architecture
+Unidirectional Peggy focuses on core features for unidirectional transfers. This prototype includes functionality to safely lock and unlock Ethereum and ERC20 tokens, emitting associated events which provide the Cosmos based Relayer and Oracle modules with the information required to complete unidirectional transfers. Validators witness the events and submit proof in the form of signed hashes to the Cosmos based modules, which are responsible for aggregating and tallying the Validators’ signatures and their respective signing power. The system is managed by the contract's deployer, designated internally as the relayer, a trusted third-party which can unlock funds and return them their original sender. If the contract’s balances under threat, the relayer can pause the system, temporarily preventing users from depositing additional funds.
```
cwgoes: What would it mean for the contract's balances to be under threat?

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
...
-    - Open Ganache app, open the *Settings* button and make sure it's running on port **_7545_**.
-    - `$ truffle test --network ganache` on a separate tab
+The Peggy Smart Contract is deployed on the Ropsten testnet at address: 0x3de4ef81Ba6243A60B0a32d3BCeD4173b6EA02bb
 
+### Architecture Diagram
+![peggyarchitecturediagram](https://user-images.githubusercontent.com/15370712/58388886-632c7700-7fd9-11e9-962e-4e5e9d92c275.png)
 
-![alt text](./img/ganache_setup.png "Ganache Setup")
+### System Process:
+1. Users lock Ethereum or ERC20 tokens on the Peggy contract, resulting in the emission of an event containing the created item's original sender's Ethereum address, the intended recipient's Cosmos address, the type of token, the amount locked, and the item's unique nonce.
```
cwgoes: "created item" is not clear - what "item" is being referred to? What is the nonce? Is this a sort of packet?

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
...
+2. Cosmos based Validators witness these lock events via a Relayer module and sign a hash containing the unique item's information, which is relayed from the Cosmos Relayer module to the Oracle module.
```
cwgoes:
```suggestion
2. Validators on the Cosmos chain witness these lock events via a Relayer module and sign a hash containing the unique item's information, which is relayed from the Cosmos Relayer module to the Oracle module.
```

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
...
+3. Once the Oracle module has verified that the validators' aggregated signing power is greater than the specified threshold, it mints the appropriate amount of tokens and forwards them to the intended recipient.
```
cwgoes: Is the idea here that the Relayer and Oracle module exist on the same chain? Would be helpful to explain that separation more clearly.

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
...
-       type: 'someType',
-       event: 'SomeEvent',
-       args: [Object] } ] }
-  ```
+## Security, Privacy, Risks
+Disclaimer: These contracts are for testing purposes only and are NOT intended for production. In order to prevent any loss of user funds, locked Ethereum and ERC20 tokens can be withdrawn directly by the original sender at any time. However, these contracts have not undergone external audits and should not be trusted with mainnet funds. Any use of Peggy is at the user’s own risk.
 
-Copy it and on in a separate tab in the console:
+## Other Considerations
+We decided to temporarily remove the validator set from this version of Peggy, our reasoning being that system transparency and clarity should take precedence over the inclusion of future project features.
```
cwgoes: What do you mean? That there still is *a* validator set - just not the same as the Tendermint validator set?

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
...
+## Project Background
+We are hoping to create a closed system for intra network transfers of cryptocurrency between blockchains, spearheaded by a proof-of-concept which enables secured transactions between Ethereum and Cosmos.
 
-## Instructions
+## Smart Contract Scope
 
-You can test the contracts using Truffle Testrpc in the console or Ganache UI.
+### Goals of the Smart Contracts
+1. Securely implement core functionality of the system such as asset locking and event emission without endangering any user funds. As such, this prototype does not permanently lock value and allows the original sender full access to their funds at any time.
+2. Interface with the Relayer service, which is used by validators to listen for contract events which are signed and submitted to the Cosmos network as proof of transaction.
```
cwgoes: Is the "Relayer" a service or a module (or both)? It's called both in different places in this file.

ethereum-contracts/README.md
```diff
@@ -1,72 +1,61 @@
-# Testing contracts
+# Unidirectional Peggy Project Specification
 
-## Installing dependencies
+## Project Summary
+Unidirectional Peggy is the starting point for cross chain value transfers from Ethereum to Cosmos as part of the Ethereum Cosmos Bridge project. The smart contract system accepts incoming transfers of Ethereum and ERC20 tokens, locking them while the transaction is validated and equitable funds issued to the intended recipient on Cosmos.
```
cwgoes: Can this part clarify that "Cosmos" is a specific bridge chain running Tendermint consensus and custom modules?

By itself, "Cosmos" refers to the network as a whole, not a particular blockchain.

Or is this intended to become an upgrade for the Cosmos Hub (which is a particular blockchain)?

ethereum-contracts/README.md
denalimarsh: Correct. The original sender can withdraw tokens using `withdraw(bytes32 id)` while the contract's relayer can return tokens to the item's original sender using `unlock(id)`. This is intended behavior for this stage of the project to protect any user funds should someone attempt to deploy the contracts to the mainnet; once the bridge has undergone robust testing the `withdraw()` function will be removed entirely and the `unlock()` function will be contingent on approval from the validator set.

I've updated the project specification to state that this project specification focuses on the role of 'Peggy', the Smart Contract system deployed to the Ethereum network, and is meant to contextualize its role within the bridge. Specifications detailing the structure and step=-by-step process of the non-Ethereum components (Relayer service, EthOracle module, Oracle module) will soon be available in the cosmos-ethereum-bridge repository.

ethereum-contracts/README.md
denalimarsh: 'Under threat' refers to any unforeseen circumstances which endanger balances on the contract, such as code insecurity. Pausing locking is a security measure which prevents additional funds from being deposited and therefore at risk in such a scenario.

ethereum-contracts/README.md
denalimarsh: Updated.

ethereum-contracts/README.md
denalimarsh: Updated with: 'The Relayer is a service which interfaces with both blockchains, allowing validators to attest on the Cosmos blockchain that specific events on the Ethereum blockchain have occurred. The Relayer listens for `LogLock` events, parses information associated with the Ethereum transaction, uses it to build unsigned Cosmos transactions, and enables validators to sign and send the transactions to the Oracle module on Cosmos.'

ethereum-contracts/README.md
denalimarsh: Correct. This refers to the validator set from Valset.sol, which was removed from the Smart Contracts for the unidirectional implementation. I've updated the section to be more explicit.

ethereum-contracts/README.md
denalimarsh: The Relayer is a service, apologies for the confusion.

ethereum-contracts/README.md
denalimarsh: I believe that the custom modules will eventually be submitted to the Cosmos sdk as an upgrade for the Cosmos Hub and other blockchains Cosmos network. @Jazzy and @jackzampolin do you have any insight on this?

ethereum-contracts/README.md
denalimarsh: Yes, an Item is a struct containing information about a unique incoming transaction which executes the `lock()` function. Due to confusion about the name, we should rename the struct to something more descriptive. How about 'LockedTokens' or 'Deposit'?


The nonce is a global variable which protects Peggy against possible replay attacks so that an item's funds can only be unlocked once. Consider this attack scenario:
- An adversary locks X ETH into the contract (move from Ethereum to Cosmos Zone).
- He then requests the reverse operation (move from Cosmos Zone to Ethereum).
- After collecting the validator signatures, the relayer creates a transaction calling the `unlock` function for the amount X to sent to the user.
- The adversary observes the transaction on the blockchain, and copies the signatures
- The adversary, having already received X ETH, creates another transaction calling `unlock` directly himself for the amount X and replaying the captured signatures.
- The amount X is transferred to the adversary again
- The attacker can repeat this attack until the contract has been drained of ETH.

ethereum-contracts/contracts/Peggy.sol
denalimarsh: Items aren't 1:1 with Msgs, as Items serve as storage for the original sender's funds and include a unique id which isn't currently reflected in the Msg sent on Cosmos - but I agree that 'Msg' is a much more accurate name than 'Item'. @cwgoes what are your thoughts on the naming conventions?

ethereum-contracts/README.md
cwgoes: `Deposit` sounds great - and a nonce makes sense for replay prevention, agreed.

ethereum-contracts/README.md
cwgoes: OK.

denalimarsh:
Future requests for revisions should be directed to #51 and any outstanding comments from this PR will be addressed in #51 . This PR will be removed once the relevant revisions have been integrated in the completed unidirectional branch.

denalimarsh:
Closing this as #51 contains the up-to-date version of the smart contracts which include any requested changes from this PR.

PR #63, opened 2019-09-16
feat/contract-tooling

denalimarsh:
This PR enhances Peggy smart contract interaction by integrating truffle into the project's development environment and usage.

- [x] Importing OpenZeppelin contracts directly from openzeppelin-solidity library
- [x] Local end-to-end testing with truffle: deploy peggy contract, ebrelayer connection, send lock transaction to peggy contract
- [x] Yarn workspaces and user-friendly commands: `yarn peggy:abi`, `yarn peggy:address`, `yarn:peggy lock`, etc.
- [x] Ropsten end-to-end test with Infura and MetaMask
- [x] Updated README with all commands/steps for end-to-end test

This PR has been preliminarily reviewed in the swishlabsco/peggy repo: https://github.com/swishlabsco/peggy/pull/2.

***Lock transaction on local truffle blockchain***
Screen Shot 2019-09-13 at 12 07 09 PM

***ebrelayer witnesses LogLock event***
Screen Shot 2019-09-13 at 12 07 39 PM

_Future work:_
tmux bash script which executes a local end-to-end test

PR #68, opened 2019-09-25
feat/tag-chain-and-contract

denalimarsh:
fixes #58 

- [x] Added ChainID, BridgeContractAddress, Symbol, and TokenContractAddress to EthBridgeClaim
- [x] Support for the new fields in LockEvent, including symbol data fetching by Peggy smart contract
- [x] Added new fields to ethbridge cli, querier
- [x] Updated the README with examples that reflect the new parameters
- [x] Successful compilation and testing of cli commands

This feature has been preliminarily reviewed in the swishlabs/peggy repo https://github.com/swishlabsco/peggy/pull/4

***tx create-claim, query prophecy***
Screen Shot 2019-09-25 at 12 23 42 AM

This feature is built on top of https://github.com/cosmos/peggy/pull/64. Once https://github.com/cosmos/peggy/pull/63 and https://github.com/cosmos/peggy/pull/65 have been approved, this feature will be rebased to the updated sdk and tested with the new contract tools.

PR #71, opened 2019-10-04
Dockerize validator and relayer for simple testing purposes

musnit:
This adds a docker and docker-compose that runs the validator and relayer together with non-secret private keys that are expected to be used only for local testing or for a public-facing testnet intended for developer testing only.

There will be a follow up PR that adds README and Makefile commands for instructions on how to build and deploy this to AWS.

It is currently blocked on https://github.com/cosmos/peggy/pull/63, and contains all important commits from that branch. We can deprecate that branch if desired, or merge that branch first and then this one second.


cmd/ebrelayer/contract/contract_test.go
```diff
@@ -12,11 +9,10 @@ import (
 // TestLoadABI : test that contract containing named event is successfully loaded
 func TestLoadABI(t *testing.T) {
 
+	const AbiPath = "/src/github.com/cosmos/peggy/cmd/ebrelayer/contract/abi/Peggy.abi"
```
golangcibot: const `AbiPath` is unused (from `unused`)

musnit: This is fixed in follow up PR

musnit:
Note: Have retargeted this to cosmos:feat/tag-chain-and-contract and it is also blocked on https://github.com/cosmos/peggy/pull/68

denalimarsh:
@musnit resolved a merge conflict. As #63 and #68 have been merged, let's change the base branch back to master and resolve any conflicts so that we can merge this PR.

musnit:
@denalimarsh @fedekunze have rebased this onto master.

cmd/ebrelayer/main.go
```diff
@@ -77,7 +78,7 @@ func initRelayerCmd() *cobra.Command {
 	initRelayerCmd := &cobra.Command{
 		Use:   "init [web3Provider] [contractAddress] [eventSignature] [validatorFromName] --chain-id [chain-id]",
 		Short: "Initializes a web socket which streams live events from a smart contract",
-		Args:  cobra.ExactArgs(4),
+		Args:  cobra.MinimumNArgs(4),
```
fedekunze: undo, we should always provide exact args. If there's an optional arg we can use a flag instead for it. We'd also need to add a `Long:` description to provide a concrete example for using the flag.

musnit:
@fedekunze have fixed args/flag and rebased to master. should be ready for merge!

PR #74, opened 2019-10-08
Update to use new supply module

musnit:
As part of the task to add burning to burn cEth to be sent back to ether, I've updated to use the supply module.

I'm not 100% sure it was integrated properly, as the docs on it are still limited but I think it's good and it all works well. Let me know if any issues, thanks!

### DIFF
--- a/x/ethbridge/test_common.go
+++ b/x/ethbridge/test_common.go
@@ -1,0 +1,24 @@
+package ethbridge
+
+import (
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/cosmos-sdk/x/bank"
+	"github.com/cosmos/cosmos-sdk/x/supply"
+	"github.com/cosmos/peggy/x/ethbridge/types"
+	oracle "github.com/cosmos/peggy/x/oracle"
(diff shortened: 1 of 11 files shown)


musnit:
@fedekunze updated to base off master

x/ethbridge/test_common.go
```diff
@@ -0,0 +1,24 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/cosmos/cosmos-sdk/x/bank"
+	"github.com/cosmos/cosmos-sdk/x/supply"
+	"github.com/cosmos/peggy/x/ethbridge/types"
+	oracle "github.com/cosmos/peggy/x/oracle"
+	keeperLib "github.com/cosmos/peggy/x/oracle/keeper"
+)
+
+func CreateTestHandler(t *testing.T, consensusNeeded float64, validatorAmounts []int64) (sdk.Context, oracle.Keeper, bank.Keeper, supply.Keeper, []sdk.ValAddress, sdk.Handler) {
```
fedekunze: nit, it'd be better to create a test suite for this: https://godoc.org/github.com/stretchr/testify/suite

fedekunze: it'll allow you to also rename this file to `common_test.go`

PR #75, opened 2019-10-08
Add messages to burn coins through the bridge and emit events

musnit:
Note: This is a draft/WIP in PR.

This PR adds a message for burning coins through the bridge module which then emits events that can be detected and dealt with by the relayer. The relayer and solidity code will be in a future PR.

 - It is blocked on https://github.com/cosmos/peggy/pull/68 and will likely change a bit once that is merged. That should ideally be merged first
 - It is also blocked on https://github.com/cosmos/peggy/pull/74, that should be merged first. This PR contains commits from 74 so will become smaller once that is merged.
 - We should add some tests around nonce, or possibly remove nonce and just use transaction hash to prevent duplicate burns from being possible on the ethereum side of the bridge.
- I also need to update the README with burn related instructions


jackzampolin:
@musnit Don't know how that would happen (re: events) if the module adds the tags to the transaction properly. You should see them...

PR #76, opened 2019-10-10
Contract: cosmos event processing, validator claims, and token minting

denalimarsh:
This PR expands the smart contracts to support the processing of Cosmos based events, including the core functionality for validator signature verification and consensus.

- [x] Enables claims on Cosmos network transactions
- [x] Enables validators to make Oracle claims on Cosmos network transaction claims
- [x] Implements validator signature verification
- [x] Enables the processing of prophecies on a collection of Oracle claims via validator consensus
- [x] Minting of tokens on-chain in the event of a successful prophecy
- [x] Test suite for the new contracts: CosmosBridge, Oracle, Bank. The test suite can be run with `truffle test` from the `testnet-contracts` dir.

***Successful run of test suite***
Screen Shot 2019-10-08 at 5 03 51 PM

This PR is currently blocked by https://github.com/cosmos/peggy/pull/63, which contains base functionality required by this branch. For ease of review, this PR has been made against https://github.com/cosmos/peggy/pull/63 instead of master in order to declutter the commit history.

Thanks to @mossid, @fedekunze, and @adrianbrink for their past work on validator signature verification and consensus.

Considerations and ongoing refactoring:
- Refactor `Recoverer.sol` to OpenZeppelin's `ECDSA.sol` library
- Consider refactoring `pragma experimental ABIEncoderV2`
- Resolve the gas deployment issue introduced by Bank method `deployBankToken()`
- Add Peggy functions `makeOracleClaimOnCosmosBridgeClaim()` and `processProphecyOnOracleClaims()` to the test suite
- Update name of 'cosmosBridgeNonce' to 'cosmosBridgeClaimId'


testnet-contracts/contracts/BankToken.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.0;
+
+import "../../node_modules/openzeppelin-solidity/contracts/token/ERC20/ERC20Mintable.sol";
+
+/**
+ * @title BankToken
+ * @dev Mintable ERC20 token controlled by bank contract
+ **/
+
+contract BankToken is ERC20Mintable {
```
musnit: Do we want to use ERC20Detailed here too? And then just make name == _symbol and use the inherited constructor?

testnet-contracts/contracts/BankToken.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity ^0.5.0;
+
+import "../../node_modules/openzeppelin-solidity/contracts/token/ERC20/ERC20Mintable.sol";
+
+/**
+ * @title BankToken
```
musnit: Thinking about naming, this is basically a generic wrapper around ERC20Detailed and ERC20Mintable, so maybe just call it ERC20DetailedMintable?

testnet-contracts/contracts/Bank.sol
```diff
@@ -0,0 +1,96 @@
...
+
+import "../../node_modules/openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "./BankToken.sol";
+
+/**
+ * @title Bank
+ * @dev Manages the deployment and minting of ERC20 compatible BankTokens
+ **/
+
+contract Bank {
```
musnit: Thinking about naming, maybe should be BridgeBank?

testnet-contracts/contracts/Bank.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    using SafeMath for uint256;
+
+    mapping(address => bool) public bankTokens;
+    uint256 public numbTokens;
```
musnit: numbTokens...i read that as numb tokens. Maybe rather call it numTokens?

testnet-contracts/contracts/Bank.sol
```diff
@@ -0,0 +1,96 @@
...
+     */
+     function deliver(
+        address _token,
+        string memory _symbol,
+        uint256 _amount,
+        address _beneficiary
+    )
+        internal
+    {
+        // If no token address, deploy a new bank token
```
musnit: Hmm I think it may be better to make deployment happen outside of the deliver method rather than inside, ie, deliver always fails if the token has not been deployed yet?

testnet-contracts/contracts/Bank.sol
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @param _symbol: bank token symbol
+     */
+    function deployBankToken(
+        string memory _symbol
+    )
+        internal
+        returns(address)
+    {
+        numbTokens = numbTokens.add(1);
```
musnit: numTokens

testnet-contracts/contracts/Bank.sol
```diff
@@ -0,0 +1,96 @@
...
+
+        // TODO: BankToken contract deployment puts Peggy over gas limit, causing deployment to fail
```
musnit: ERC20 deployment is pretty expensive then? As mentioned before, putting it in a separate transaction may be best then. Also, thinking about optimising Peggy. lets discuss on call.

testnet-contracts/contracts/CosmosBridge.sol
```diff
@@ -0,0 +1,131 @@
...
+    using SafeMath for uint256;
+
+    /*
+    * @dev: Public variable declarations
+    */
+    uint256 public cosmosBridgeNonce;
+    mapping(uint256 => CosmosBridgeClaim) public cosmosBridgeClaims;
+
+    enum Status {
+        Active,
```
musnit: can we keep this in sync with the Go implementation, ie, pending, success and failed

testnet-contracts/contracts/CosmosBridge.sol
```diff
@@ -0,0 +1,131 @@
...
+        address _validatorAddress,
+        address _tokenAddress,
+        string _symbol,
+        uint256 _amount
+    );
+
+    /*
+    * @dev: Modifier to restrict access to completed CosmosBridgeClaims
+    */
+    modifier isProcessing(
```
musnit: isPending

testnet-contracts/contracts/Recoverer.sol
```diff
@@ -0,0 +1,21 @@
+pragma solidity ^0.5.0;
+
+// TODO: Refactor Recoverer to ECDSA based signature validation
+contract Recoverer {
```
musnit: Is there an OpenZeppelin or widely used version of this contract that we can use for functionality rather than implement our own?

testnet-contracts/contracts/Oracle.sol
```diff
@@ -0,0 +1,297 @@
+pragma solidity ^0.5.0;
+pragma experimental ABIEncoderV2;
+
+import "../../node_modules/openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "./Recoverer.sol";
+
+contract Oracle is Recoverer {
```
musnit: Should Recoverer be a library rather than something we inherit?

testnet-contracts/contracts/Oracle.sol
```diff
@@ -0,0 +1,297 @@
...
+    // Maps CosmosBridgeClaim id to OracleClaims made on it by validators
+    mapping(uint256 => OracleClaim[]) public oracleClaims;
+    // Maps validator's address to each CosmosBridgeClaim they've made
+    mapping(address => uint256[]) public validatorOracleClaims;
+
+    /*
+    * @dev: Temporary buffer which records address use in a processProphecyClaim attempt,
+    *       since the Solidity compiler does not yet implement memory (local) mappings
+    */
+    mapping(address => bool) internal usedAddresses;
```
musnit: Using storage is one of the most gas intensive operations. I'm not sure it makes sense to use this when we only need it for memory purposes, and we likely want to consider an alternative data structure/algorithm logic for doing the usedAddresses check even if it means worse-looking code. Can discuss on call

testnet-contracts/contracts/Oracle.sol
```diff
@@ -0,0 +1,297 @@
...
+import "./Recoverer.sol";
+
+contract Oracle is Recoverer {
+
+    using SafeMath for uint256;
+
+    /*
+    * @dev: Public variable declarations
+    */
+    address public provider;
```
musnit: Provider is confusing as its a defined term in the web3 space with web3Provider. Should we use operator rather?

testnet-contracts/contracts/Oracle.sol
```diff
@@ -0,0 +1,297 @@
...
+            totalPower,
+            _submitter
+        );
+    }
+
+    /*
+    * @dev: updateValidatorsPower
+    *       Allows the provider to update both the validators and powers
+    */
+   function updateValidatorsPower(
```
musnit: We likely want methods to make smaller updates too that are less gas intenstive, like, to update/add/remove a single validator.

testnet-contracts/contracts/Peggy.sol
```diff
@@ -1,18 +1,20 @@
...
-   *  @dev: Peg zone contract for testing one-way transfers from Ethereum
-   *        to Cosmos, facilitated by a trusted provider. This contract is
-   *        NOT intended to be used in production and users are empowered
-   *        to withdraw their locked funds at any time.
+   *  @dev: Peg zone contract two-way asset transfers between Ethereum and
+   *        to Cosmos, facilitated by a set of validators. This contract is
+   *        NOT intended to be used in production (yet).
    */
-contract Peggy is Processor {
+contract Peggy is CosmosBridge, Oracle, Bank, Processor {
```
musnit: I think we may want to align the solidity implementation with the go implementation. Specifically what comes to mind is that in the Go implementation we have something like:
 - transaction first goes to bridge module
 - bridge module *uses* oracle module
 - oracle module returns result to bridge module
 - bridge module processes based on result

compared to this implementation, where we have a Peggy contract that is inheriting all the above functionality and state and is this single massive context doing everything.

Are there any issues/concerns/gas inefficiencies with the componentization/code abstraction based on the Go version that are not applicable or non-standard in Solidity, or would the Go abstractions work fine? Can discuss on call.

testnet-contracts/test/test_oracle.js
```diff
@@ -0,0 +1,386 @@
...
+          [userOne, userTwo, userTwo],
+          sigV,
+          sigR,
+          sigS
+        )
+        .should.be.rejectedWith("does not meet the threshold");
+    });
+  });
+});
+
```
musnit: need test+functionality for failed prophecies where consensus becomes impossible

testnet-contracts/contracts/Peggy.sol
```diff
@@ -138,55 +136,90 @@ contract Peggy is Processor {
...
-            amount,
-            uniqueNonce
+    /*
+    * @dev: processProphecyOnOracleClaims
+    *       Processes an attempted prophecy on a CosmosBridgeClaim's OracleClaims
+    */
+   function processProphecyOnOracleClaims(
+        uint256 _cosmosBridgeNonce,
+        bytes32 _hash,
+        address[] memory _signers,
```
musnit: looks like we're expecting all signatures to be gathered offchain and posted together in a single transaction? this is fine process-wise, but not sure if fits our current relayer implementation and plans for V1, so lets discuss on call.

testnet-contracts/contracts/Peggy.sol
denalimarsh: As of v0.5.0, "Specifying base constructor arguments multiple times in the same inheritance hierarchy is now disallowed". Ideally the Bridge contract inherits from the Oracle contract which inherits from a Valset contract. However, since our Valset is deployed with constructor arguments `initValidatorAddresses` and `initValidatorPowers` we can only pass them once through the inheritance hierarchy. Hence the subsequent decision to combine Valset + Oracle and have Peggy inherit from it directly.

There are multiple other potential solutions (no constructor arguments just call ```updateValidatorsPower(newValidators, newPowers)```, or deploy the Valset as an isolated contract, for example) which we can discuss.

testnet-contracts/contracts/Recoverer.sol
denalimarsh: Yes, available for v0.5.0 as a library: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/cryptography/ECDSA.sol. Will complete the TODO by swapping it in for the Recoverer contract.

testnet-contracts/contracts/Oracle.sol
denalimarsh: Yes, see comment above.

testnet-contracts/contracts/Bank.sol
denalimarsh: Renamed this to CosmosBank in https://github.com/cosmos/peggy/pull/76/commits/89cb10e8d06ff3dcd467bcfc7b2821bf14240153 (with Processor renamed to EthereumBank for clarity). Let's discuss an architecture along the lines of: EthereumBank, CosmosBank -> BridgeBank -> Peggy.

testnet-contracts/contracts/BankToken.sol
denalimarsh: Renamed to CosmosToken in https://github.com/cosmos/peggy/pull/76/commits/89cb10e8d06ff3dcd467bcfc7b2821bf14240153. I think Bridge token is better as it is more generic and will update the naming conventions again.

testnet-contracts/contracts/Bank.sol
denalimarsh: Updated to ```cosmosTokenCount``` in https://github.com/cosmos/peggy/pull/76/commits/89cb10e8d06ff3dcd467bcfc7b2821bf14240153

testnet-contracts/contracts/Bank.sol
denalimarsh: We could deploy Peggy across 2 blocks, but I think the proposed solution of a separate transaction is preferable.

denalimarsh:
> Overall I think we should call and chat through them all, and discuss general considerations around:
> 
> * Naming and abstractions, specifically with regards to matching the Go implementation
> * Gas issues and optimizations
> * Process for how we expect this contract to be used by the relayer service and how it will call methods

@musnit thanks for the review! There are some v0.5.0 Solidity specific caveats for an implementation which matches the existing Go implementation exactly (described above). Looking forward to discussing how the relayer service will interact with the contracts and designing/optimizing within those constraints.

testnet-contracts/test/test_oracle.js
denalimarsh: These tests will be included in the upcoming dynamic Valset PR, as consensus becoming impossible is linked to changes in the validator set.

PR #82, opened 2019-10-16
Valset, CosmosBridge, Oracle, and BridgeBank

denalimarsh:
This PR refactors and expands the smart contracts into four distinct entities:
1. **Valset**: manages validator set and validator signature verification
2. **CosmosBridge**: manages ProphecyClaims, which are initial validator submissions of events that have occurred on another blockchain
3. **Oracle**: manages OracleClaims, which are additional validator submissions which confirm that the original ProphecyClaim is accurate. Manages the processing of prophecies, which are attempts at completing a ProphecyClaim using the OracleClaims made upon it.
4. **BridgeBank**: manages BridgeTokens, which assets controlled according to the bridge's rules
a. **CosmosBank**: deploys and mints BridgeTokens which represent assets on Cosmos
b. **EthereumBank**: locks and unlocks BridgeTokens which represent assets on Ethereum

The system is managed by an operator which oversees the deployed contracts.

**1. Valset tests**
```
  Contract: Valset
    Valset contract deployment
      ✓ should deploy the Valset and correctly set initial validator count
      ✓ should correctly set initial validators (47ms)
      ✓ should correctly set initial validator powers  (49ms)
      ✓ should correctly set the total power
    Signature verification
      ✓ should correctly validate signatures
      ✓ should not validate signatures on a different hashed message
      ✓ should not validate signatures from a different address
```

**2. CosmosBridge tests**
```
Contract: CosmosBridge
    CosmosBridge smart contract deployment
      ✓ should deploy the CosmosBridge with the correct parameters
      ✓ should allow the operator to set the Oracle (51ms)
      ✓ should not allow the operater to update the Oracle once it has been set (75ms)
      ✓ should allow the operator to set the Bridge Bank (51ms)
      ✓ should not allow the operater to update the Bridge Bank once it has been set (74ms)
    Creation of bridge claims
      ✓ should allow for the creation of new bridge claims (58ms)
      ✓ should log an event containing the new bridge claim's information (59ms)
      ✓ should increase the bridge claim count upon the creation of new a bridge claim (85ms)
    Bridge claim status
      ✓ should not show fake bridge claims as active
      ✓ should allow users to check if a bridge claim is currently active (69ms)
      ✓ should allow users to check if a bridge claim's original validator is currently an active validator (76ms)
```

**3. Oracle tests**
```
  Contract: Oracle
    Oracle smart contract deployment
      ✓ should deploy the Oracle, correctly setting the operator and valset
    Creation of oracle claims
      ✓ should not allow oracle claims upon inactive bridge claims (51ms)
      ✓ should not allow non-validators to make oracle claims (42ms)
      ✓ should not allow validators to make OracleClaims with invalid signatures (70ms)
      ✓ should not allow validators to make OracleClaims with another validator's signature (62ms)
      ✓ should allow validators to make OracleClaims with their own signatures (58ms)
      ✓ should not allow validators to make duplicate OracleClaims (110ms)
      ✓ should emit an event containing the new OracleClaim's information (54ms)
    Prophecy processing
      ✓ should not process the prophecy if signed power does not pass the required threshold power (102ms)
      ✓ should allow for the processing of prophecies (226ms)
      ✓ should allow non-unanimous consensus if signed power passes threshold (240ms)
      ✓ should not allow a prophecy to be processed twice (214ms)
      ✓ should emit an event upon successful prophecy processing (258ms)
```

**4. BridgeBank tests**
```
  Contract: BridgeBank
    BridgeBank deployment and basics
      ✓ should deploy the BridgeBank, correctly setting the operator and valset
      ✓ should correctly set initial values of CosmosBank and EthereumBank (43ms)
      ✓ should not allow a user to send ethereum directly to the contract
    BridgeToken creation (Cosmos assets)
      ✓ should not allow non-operators to create new bridge tokens (38ms)
      ✓ should allow the operator to create new bridge token (51ms)
      ✓ should emit event LogNewBridgeToken containing the new bridge token's address and symbol (73ms)
      ✓ should increase the bridge token count upon creation (81ms)
      ✓ should add the new bridge token to the whitelist (88ms)
      ✓ should allow the creation of bridge tokens with the same symbol (200ms)
    Bridge token minting (Cosmos assets)
      ✓ should mint bridge tokens upon the successful processing of a prophecy claim (158ms)
    Bridge token deposit locking (Ethereum/ERC20 assets)
      ✓ should allow users to lock ERC20 tokens (88ms)
      ✓ should allow users to lock Ethereum (53ms)
      ✓ should generate unique deposit ID for a new deposit (38ms)
      ✓ should correctly mark new deposits as locked (103ms)
      ✓ should be able to access the deposit's information by its ID (104ms)
      ✓ should correctly store deposit information (121ms)
    Bridge token deposit unlocking (Ethereum/ERC20 assets)
      ✓ should allow for an Ethereum deposit to be unlocked (41ms)
      ✓ should allow for an ERC20 deposit to be unlocked (47ms)
      ✓ should not allow for the unlocking of non-existant deposits (38ms)
      ✓ should not allow an unlocked deposit to be unlocked again (76ms)
      ✓ should update lock status of deposits upon completion (74ms)
      ✓ should emit an event upon unlock with the correct deposit information (45ms)
      ✓ should correctly transfer unlocked Ethereum (53ms)
      ✓ should correctly transfer unlocked ERC20 tokens (120ms)
```

Some features contained in this PR have undergone a preliminary review in https://github.com/cosmos/peggy/pull/76.

Thanks to @mossid, @fedekunze, and @adrianbrink for their past work on validator signature verification and consensus.


testnet-contracts/contracts/Oracle.sol
```diff
@@ -0,0 +1,243 @@
...
+        emit LogProphecyProcessed(
+            _bridgeClaimID,
+            weightedSignedPower,
+            weightedTotalPower,
+            msg.sender
+        );
+    }
+
+    /*
+    * @dev: processProphecyClaim
```
musnit: should be checkProphecyClaim?

testnet-contracts/contracts/CosmosBridge.sol
```diff
@@ -0,0 +1,295 @@
...
+        bytes _cosmosSender,
+        address payable _ethereumReceiver,
+        address _validatorAddress,
+        address _tokenAddress,
+        string _symbol,
+        uint256 _amount
+    );
+
+    event LogBridgeClaimCompleted(
+        uint256 _bridgeClaimID
```
musnit: I think bridgeClaimID should actually be bridgeProphecyID, right? implementation looks fine, i'm just refering to naming...and similarly, related functions should be like processProphecy instead of processClaim?

testnet-contracts/contracts/Oracle.sol
```diff
@@ -0,0 +1,243 @@
...
+
+    // Tracks the number of OracleClaims made on an individual BridgeClaim
+    mapping(uint256 => address[]) public oracleClaimValidators;
+    mapping(uint256 => mapping(address => bool)) public hasMadeClaim;
+
+    /*
+    * @dev: Event declarations
+    */
+    event LogNewOracleClaim(
+        uint256 _bridgeClaimID,
```
musnit: Again, we're kind of mixing the concept of _bridgeClaimID and _prophecyID. They are the same thing conceptually and I think we can just call it bridgeProphecyID everywhere?

testnet-contracts/contracts/Oracle.sol
```diff
@@ -0,0 +1,243 @@
...
+            ) == true,
+            "Can only check active bridge claims"
+        );
+        return processClaim(
+            _bridgeClaimID
+        );
+    }
+
+    /*
+    * @dev: processClaim
```
musnit: hmm the term process to me implies that some completion/mutation may have occured, so could be a little ambiguous. maybe a better name for this, like, checkProphecyThreshold ?

testnet-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -0,0 +1,271 @@
...
+            _amount
+        );
+    }
+
+    /*
+     * @dev: Unlocks Ethereum deposits.
+     *
+     * @param _id: Unique key of the CosmosDeposit.
+     */
+    function unlock(
```
musnit: is this function just for the operator, as a backdoor? its not really needed/part of the end-goal production version right?

it also looks like there is no implementation yet of unlocking ethereum/erc20 tokens to be sent out to a receiver under the typical bridge flow of claims/prophecies - is that missing? specifically, i mean the flow of burning cETH on cosmos with a target destination ethereum address, then waiting for relayers/validators to make the claims and final prophecy that unlocks that real ETH and sends it to the destination chosen during the initial burn.

maybe these two flows are being confused here

testnet-contracts/contracts/CosmosBridge.sol
```diff
@@ -0,0 +1,295 @@
...
+
+        hasBridgeBank = true;
+        bridgeBank = BridgeBank(_bridgeBank);
+
+        emit LogBridgeBankSet(
+            address(bridgeBank)
+        );
+    }
+
+    // TODO: BridgeClaims can only be created for BridgeTokens on BridgeBank's whitelist.
```
musnit: Remove TODO, make this a normal comment. We probably need to document this whitelisting step in the README more clearly too which can be done in follow up PR. The current implementation will then result in unfinalized prophecies left dangling until the token has been whitelisted, which I think could be the right choice as it will allow easy checks for this situation, allowing the operator to remedy it as needed when it comes up. <-- probably want to add these comments into the code comments or README too.

testnet-contracts/migrations/2_next.js
```diff
@@ -1,7 +1,46 @@
...
+const CosmosBridge = artifacts.require("CosmosBridge");
+const Oracle = artifacts.require("Oracle");
+const BridgeBank = artifacts.require("BridgeBank");
 
 module.exports = function(deployer, network, accounts) {
-  deployer.deploy(TestToken, { gas: 4612388, from: accounts[0] });
-  deployer.deploy(Peggy, { gas: 4612388, from: accounts[0] });
+  const operator = accounts[0];
+  const initialValidators = [accounts[1], accounts[2], accounts[3]];
+  const initialPowers = [5, 8, 12];
```
musnit: This migration is very specific in terms of initial validators/powers. Ideally this should be paramaterized or taken from a config file that the end user can configure themself. Maybe make our own config file or can just add some fields to the truffle config to use.

testnet-contracts/contracts/CosmosBridge.sol
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/f34f19bb8544b4e8a153af2152647283ffa8be13

testnet-contracts/contracts/Oracle.sol
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/f8edef9a6846c4f85a67968d0462354719776874

testnet-contracts/contracts/Oracle.sol
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/f8edef9a6846c4f85a67968d0462354719776874

testnet-contracts/contracts/Oracle.sol
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/60cb52d4b6baf69bc7e237ba9839b1a9e2fae1a2

testnet-contracts/contracts/CosmosBridge.sol
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/60cb52d4b6baf69bc7e237ba9839b1a9e2fae1a2

testnet-contracts/migrations/2_next.js
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/7bd1e59626ea28671a880bdadd079baa071905f4, https://github.com/cosmos/peggy/pull/82/commits/b4109a915f03e594aa0285a125f39a1070d00a46

testnet-contracts/contracts/BridgeBank/BridgeBank.sol
musnit: @denalimarsh bump?

denalimarsh: @musnit after cETH is burned on Cosmos, details of the burn event are relayed to CosmosBridge's `newProphecyClaim()` with requests for the final prophecy sent to Oracle's `processBridgeProphecy()`. If the final prophecy is successful then BridgeBank will mint tokens to the intended recipient address. This is done independently of contract funds that were on-boarded from the `lock()` function. The tokens minted by BridgeBank upon the successful processing of the final prophecy are BridgeTokens, tokens which represent the Cosmos based assets 1:1. Ethereum is also currently processed by this system, with Ethereum represented as a BridgeToken with token contract address `address(0)`.

Given this system, I propose that we update the `unlock()` function so that it swaps the sender's BridgeTokens for real Ethereum/ERC20 and burns the received BridgeTokens. If someone is holding BridgeTokens they can't redeem because nobody has locked the correlated ERC20 token on the Bridge, the BridgeTokens still have the same value since they remain redeemable 1:1. This also allows for the burning of Cosmos assets -> minting of BridgeTokens even if Bridge does not hold a balance of the ERC20 token type that the BridgeTokens represent at the time of the burn event.

musnit: "burning of Cosmos assets -> minting of BridgeTokens even if Bridge does not hold a balance of the ERC20 token type that the BridgeTokens represent at the time of the burn event" -> this would only happen if the bridge broke or the validators maliciously attacked and withdrew those assets when they weren't supposed to?

musnit: i suppose ideally as soon as cETH is burned, that swap should happen immediately and so effectively the whole burn ->mint -> swap -> burn flow becomes equivalent to the burn -> unlock flow, and the latter seems much simpler?

denalimarsh: Is it the case that assets can only be originally on-boarded to the Bridge from Ethereum? I was under the assumption that Cosmos could also be the starting point for new funds.

musnit: with the burn->unlock flow, yes.

there is also a lock-> mint flow from cosmos to eth with lockup ATOM on cosmos, mint eATOM that would support the case you mention

denalimarsh: Okay, that is indeed much simpler. I will update the contracts to reflect those asset flows :)

musnit:
@denalimarsh can we base this on master now?

testnet-contracts/contracts/BridgeBank/BridgeBank.sol
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/159cf28ee1ba897533c941aa2dfe0bbd993f29a3 includes the updated contracts and test suite

denalimarsh:
> @denalimarsh can we base this on master now?

Updated base to master and fixed conflicts.

app/app.go~HEAD
```diff
@@ -0,0 +1,245 @@
+package app
```
musnit: this file shouldn't exist. there are also a ton of commits in the commit history here that are duplicated from other PRs. you sure conflict resolution worked out properly?

denalimarsh: @musnit that file keeps getting quietly added during merges. Removed in https://github.com/cosmos/peggy/pull/82/commits/160bd8ba89149433d30891d6130a1389aee6ddd6.

testnet-contracts/migrations/2_next.js
```diff
@@ -1,7 +1,129 @@
...
+  if (network === "develop" || network === "ganache") {
+    // Operator
+    operator = accounts[0];
+    // Initial validators
+    const localValidatorCount = Number(process.env.LOCAL_VALIDATOR_COUNT);
+    if (localValidatorCount <= 0 || localValidatorCount > 9) {
+      return console.error(
+        "Must provide an initial validator count between 1-8 for local deployment."
+      );
+    } else {
```
fedekunze: nit, there's no need for the else if you are returning on the `if` above. It helps readability

testnet-contracts/migrations/2_next.js
```diff
@@ -1,7 +1,129 @@
...
+      initialValidators = accounts.slice(1, localValidatorCount + 1);
+    }
+    // Initial validator power
+    if (process.env.LOCAL_INITIAL_VALIDATOR_POWERS.length === 0) {
+      return console.error(
+        "Must provide initial local validator powers as environment variable."
+      );
+    } else {
```
fedekunze: ditto

testnet-contracts/migrations/2_next.js
```diff
@@ -1,7 +1,129 @@
...
+      initialPowers = process.env.LOCAL_INITIAL_VALIDATOR_POWERS.split(",");
+    }
+    // Testnet/mainnet network deployment param parsing/setting
+  } else {
```
fedekunze: ditto

testnet-contracts/migrations/2_next.js
```diff
@@ -1,7 +1,129 @@
...
+    // Operator
+    if (process.env.OPERATOR.length === 0) {
+      return console.error(
+        "Must provide operator address as environment variable."
+      );
+    } else {
```
fedekunze: ditto

testnet-contracts/migrations/2_next.js
```diff
@@ -1,7 +1,129 @@
...
+  // 0. Deploy BridgeToken for testing purposes
+  deployer.deploy(BridgeToken, "TEST", { gas: 4612388, from: operator });
+
+  // 1. Deploy Valset contract
+  deployer
+    .deploy(Valset, operator, initialValidators, initialPowers, {
+      gas: 6721975, // Cost: 1,529,823
+      from: operator
+    })
+    .then(function() {
```
fedekunze: can we use `await` instead? 🙏

testnet-contracts/test/test_cosmosBridge.js
```diff
@@ -0,0 +1,356 @@
...
+
+      await this.cosmosBridge.setBridgeBank(this.bridgeBank.address, {
+        from: operator
+      }).should.be.fulfilled;
+
+      const bridgeBank = await this.cosmosBridge.bridgeBank();
+      bridgeBank.should.be.equal(this.bridgeBank.address);
+    });
+
+    it("should not allow the operater to update the Bridge Bank once it has been set", async function() {
```
fedekunze:
```suggestion
    it("should not allow the operator to update the Bridge Bank once it has been set", async function() {
```

testnet-contracts/migrations/2_next.js
fedekunze: ditto for the remaining `else` as well

testnet-contracts/migrations/2_next.js
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/683353d328fd12d6c7c71c443f5a340ebea6dd69

testnet-contracts/migrations/2_next.js
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/683353d328fd12d6c7c71c443f5a340ebea6dd69

testnet-contracts/migrations/2_next.js
denalimarsh: Need this `else` for input validation on non-local networks, hopefully the updated formatting in https://github.com/cosmos/peggy/pull/82/commits/683353d328fd12d6c7c71c443f5a340ebea6dd69 makes this less confusing

testnet-contracts/migrations/2_next.js
denalimarsh: https://github.com/cosmos/peggy/pull/82/commits/683353d328fd12d6c7c71c443f5a340ebea6dd69

testnet-contracts/migrations/2_next.js
denalimarsh: Updated migrations script with await/async in https://github.com/cosmos/peggy/pull/82/commits/ed748c7f8ed2bd377feb8be11969e76fde0f0773. Also added some more comments with all the gas usage/deployment costs for each contract.

testnet-contracts/test/test_cosmosBridge.js
denalimarsh: Fixed in https://github.com/cosmos/peggy/pull/82/commits/ed748c7f8ed2bd377feb8be11969e76fde0f0773

denalimarsh:
@fedekunze thanks for the review :)

I've updated the migration script's formatting/comments and added this PR to the CHANGELOG.

Issue #84, opened 2019-10-25
relayer ropsten network integration not working


arun1705:
I tried running the ropsten netwrok integration for automatic relaying process but it seems it is not working properly.The relayer is not able to connect to the deployed address showing web socket issue.I have attached screenshot for the reference
![Screenshot from 2019-10-25 11-38-03](https://user-images.githubusercontent.com/41661371/67547336-783e3b00-f71c-11e9-8803-ec7aee10be88.png)

denalimarsh:
Instead of using provider `wss://ropsten.infura.io/`, can you attempt to start the relayer with provider `wss://ropsten.infura.io/ws` as described in section 'Running the relayer service' of the README? 

Here's a screenshot which shows the issue reproduced, and then resolved by using the recommended provider:
<img width="1116" alt="Screen Shot 2019-10-24 at 11 54 58 PM" src="https://user-images.githubusercontent.com/15370712/67550064-793e8080-f6ba-11e9-95d1-47fd28b3a244.png">

Thank you for the screenshots which allowed for a quick diagnosis of this issue. Please let us know if you continue having difficulty with the relayer service.

Arunhfb:
Thank you for your quick response.Its working fine now.But i saw there is some issue with the truffle_config.js it doesnt inject the .env parameters properly.So to make things work i hard coded the parameters i.e. MNEMONIC & INFURA_PROJECT_ID into the js file directly.

And also in the below command it accepts only cosmos_recipient addresses or it can accept any other prefix addresses? 

yarn peggy:lock --network ropsten [HASHED_COSMOS_RECIPIENT_ADDRESS] [TOKEN_CONTRACT_ADDRESS] [WEI_AMOUNT]

Arunhfb:
I tried doing transaction after starting automatic relayer service but facing issues.I have attached screenshot for your reference.
![Screenshot from 2019-10-25 13-35-15](https://user-images.githubusercontent.com/52598934/67554306-bb080f00-f72c-11e9-8c17-589e845ab664.png)

denalimarsh:
I've opened PR https://github.com/cosmos/peggy/pull/85 which addresses both issues. Once it's merged, you'll be able to send the lock transaction with the following process:

1. Migrate the contracts to the ropsten network with `yarn migrate --network ropsten`
[...]

PR #77, opened 2019-10-10
Cosmos relayer

denalimarsh:
This PR introduces a Cosmos relayer which subscribes to transactions/events on the Bridge blockchain through an rpc websocket and relays the appropriate transaction to the Bridge smart contracts. For example, a `MsgBurn` event on the Bridge will result in the validator sending a tx which contains relevant event data to CosmosBridge.sol's `NewProphecyClaim()` method. 

Note: follow the steps in the README in order to configure the environment required for cosmos relayer initialization.

Example initialization cmd:
`ebrelayer init cosmos [tendermint-node] [ethereum-provider] [bridge-registry-contract-address] [validator-private-key]`. 

Example relay of `MsgBurn` event:
Screen Shot 2019-10-26 at 8 46 20 PM

This PR is blocked by https://github.com/cosmos/peggy/pull/82.

PR #85, opened 2019-10-27
Updated user scripts and infura endpoint

denalimarsh:
Hotfix which updates the Ethereum relayer scripts to be compatible with Infura's v3 endpoints and resolves https://github.com/cosmos/peggy/issues/84. These fixes have been implemented in pending PRs https://github.com/cosmos/peggy/pull/71 and https://github.com/cosmos/peggy/pull/77, but as those are still undergoing review this PR should be merged into master immediately.

### DIFF
--- a/testnet-contracts/scripts/sendLockTx.js
+++ b/testnet-contracts/scripts/sendLockTx.js
@@ -39,8 +39,10 @@ module.exports = async () => {
       );
     }
   } else if (NETWORK_ROPSTEN) {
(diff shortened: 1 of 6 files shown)


testnet-contracts/scripts/sendLockTx.js
```diff
@@ -100,14 +102,18 @@ module.exports = async () => {
   const accounts = await web3.eth.getAccounts();
```
fedekunze: side note: we should provably error if the `accounts.length === 0`

PR #90, opened 2019-11-06
update ethbridge to latest SDK

fedekunze:
- update ethbridge to latest SDK
- add PR template


app/app.go
```diff
@@ -18,16 +17,19 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/x/auth"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	"github.com/cosmos/cosmos-sdk/x/auth/vesting"
 	"github.com/cosmos/cosmos-sdk/x/bank"
-	"github.com/cosmos/cosmos-sdk/x/genaccounts"
```
musnit: so this module is removed from the sdk now, and expected to only be implemented in apps that need that command?

musnit: @fedekunze Not sure if the gov module will be needed/used, but makes sense to add it and keep in line with core sdk, thanks!

One question above, but otherwise looks good to merge.

I have not tested/confirmed everything still works fine, but am planning to do a full QA/test/fix all tests once all these PRs are merged, so we can catch anything that may have come up.

fedekunze: > Not sure if the gov module will be needed/used, but makes sense to add it and keep in line with core sdk, thanks!

Fair, I'll create an issue for it. I think we'll eventually need it for the param change proposals

fedekunze: > so this module is removed from the sdk now, and expected to only be implemented in apps that need that command?

yeah so instead of having a separate module, the genesis account logic and validation was moved to `auth`. Every new account type must expose a `Validate` function for genesis validation

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
...
+v                               ✰  Thanks for creating a PR! ✰
+v    Before smashing the submit button please review the checkboxes.
+v    If a checkbox is n/a - please still include it but + a little note why
+☺ > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > > >  -->
+
+<!-- Closes #number -->
+
+______
+
+- Targeted PR against correct branch
```
denalimarsh: should be `- [ ]`

denalimarsh: Unless this is for admin use? In which case, should be moved to that section.

Issue #150, opened 2020-03-09
Inconsistent code versus notes


okwme:
The following code seems to have an inconsistency with the notes.

```solidity
    * @dev: processProphecy
    *       Calculates the status of a prophecy. The claim is considered valid if the
    *       combined active signatory validator powers pass the validation threshold.
    *       The hardcoded threshold is (Combined signed power * 2) >= (Total power * 3).
    */
    function getProphecyThreshold(
        uint256 _prophecyID
    )
        internal
        view
        returns(bool, uint256, uint256)
    {
        uint256 signedPower = 0;
        uint256 totalPower = valset.totalPower();

        // Iterate over the signatory addresses
        for (uint256 i = 0; i < oracleClaimValidators[_prophecyID].length; i = i.add(1)) {
            address signer = oracleClaimValidators[_prophecyID][i];

                // Only add the power of active validators
                if(valset.isActiveValidator(signer)) {
                    signedPower = signedPower.add(
                        valset.getValidatorPower(
                            signer
                        )
                    );
                }
        }

        // Calculate if weighted signed power has reached threshold of weighted total power
        uint256 weightedSignedPower = signedPower.mul(3);
        uint256 weightedTotalPower = totalPower.mul(2);
        bool hasReachedThreshold = weightedSignedPower >= weightedTotalPower;
```
https://github.com/cosmos/peggy/blob/master/testnet-contracts/contracts/Oracle.sol#L195

The notes say: 
```
(Combined signed power * 2) >= (Total power * 3)
``` 
but the code says
```solidity
signedPower.mul(3) >= totalPower.mul(2)
```
Is this correct?

Issue #123, opened 2020-01-31
How to resolve incorrect lock?


okwme:
I locked an ERC-20 on Eth which was minted on SDK as Coin. Then I locked that Coin on SDK in order to return it to Eth. Of course when doing this you should not _lock_ but instead _burn_ the Coin. I did the incorrect _lock_ instead of _burn_. After adding error logging to the Eth side it throws an error that says:
```bash
> yarn peggy:process 1
yarn run v1.16.0
$ yarn workspace testnet-contracts peggy:process 1
$ truffle exec scripts/sendProcessProphecy.js 1
You can improve web3's peformance when running Node.js versions older than 10.5.0 by installing the (deprecated) scrypt package in your project
You can improve web3's peformance when running Node.js versions older than 10.5.0 by installing the (deprecated) scrypt package in your project
Using network 'ganache'.

Fetching Oracle contract...
Attempting to send processBridgeProphecy() tx...
Returned error: VM Exception while processing transaction: revert Token must be a whitelisted bridge token -- Reason given: Token must be a whitelisted bridge token.
Done in 9.56s.
```
What are the methods available for undoing an incorrectly made relay?
In a similar vein: what are the methods for relaying a transactions which was originally missed?

musnit:
This initial version of peggy is designed with the assumption that a transaction, if valid, will always succeed. This assumption should hold if a high enough threshold of  validators/relayers are all operating correctly at all times. (ie, a lock will always result in a minted coin on the opposite chain, a burn will always result in an unlock on the opposite chain)

The whitelist was added as a temporary fix, until more information/features could be gathered around having Cosmos's Coin and Bank systems support multiple tokens with the same denomination. (https://github.com/cosmos/peggy/issues/69)

So for example: in your scenario, it seems like COIN has been whitelisted, then COIN was locked on eth to create CosmosCOIN. You're trying to lock CosmosCOIN, which is not whitelisted, so it fails, preventing you from mistakenly succeeding at the incorrect lock. The way to resolve the above would to get CosmosCOIN whitelisted so that the transfer succeeds onto the Ethereum side and creates ETHCosmosCOIN, then to burn the ETHCosmosCOIN and get your CosmosCOIN back on the Cosmos side.

[...]

PR #160, opened 2020-03-20
Add consensus threshold param to oracle contract

denalimarsh:
Closes: https://github.com/cosmos/peggy/issues/150

This PR adds a consensus threshold parameter to the Oracle contract so that the threshold is configurable like it is in the oracle module.

Instead of hardcoding the threshold at `(signed power*3) >= (total power*2)`, the threshold is now set at `signed power >= x% of total power` where `x` is the consensus threshold parameter. For example, if the consensus threshold is 65, all prophecies will require 65% of the total available validator power in order to be processed.
______

For contributor use:

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [ ] Wrote unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [ ] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [ ] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Added appropriate labels to PR (ex. `WIP`, `R4R`, `docs`, etc)
- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/master/README.md#merging-a-pr))


testnet-contracts/migrations/2_next.js
```diff
@@ -81,7 +90,7 @@ module.exports = function(deployer, network, accounts) {
     // 1. Deploy BridgeToken contract
     //    Gas used:        1,884,394 Gwei
     //    Total cost:    0.03768788 Ether
-    await deployer.deploy(BridgeToken, "TEST", {
+    await deployer.deploy(BridgeToken, "stake", {
```
okwme: why is this called stake? isn't this just a dummy ERC-20 to be used as an Eth based asset that is moved to the cosmos chain?

testnet-contracts/contracts/Oracle.sol
```diff
@@ -241,15 +248,15 @@ contract Oracle {
                 }
         }
 
-        // Calculate if weighted signed power has reached threshold of weighted total power
-        uint256 weightedSignedPower = signedPower.mul(3);
-        uint256 weightedTotalPower = totalPower.mul(2);
-        bool hasReachedThreshold = weightedSignedPower >= weightedTotalPower;
+        // Prophecy must reach total signed power % threshold in order to pass consensus
```
okwme: Maybe note that the `consensusThreshold` is a decimal multiplied by 100 which is why the `signedPower` is also being multiplied by 100

testnet-contracts/migrations/2_next.js
denalimarsh: Ah, good catch. While attempting to end-to-end test around the issues with whitelisting native cosmos assets described in https://github.com/cosmos/peggy/issues/123 I updated the testing token's denom to stake. Fixed token denom in https://github.com/cosmos/peggy/pull/160/commits/b5b5d7d47d7531d2b7420dcb8da450a37bb917aa.

The whitelisting bug will be addressed within a few days as part of updating namespacing.

testnet-contracts/contracts/Oracle.sol
denalimarsh: https://github.com/cosmos/peggy/pull/160/commits/75f75ed2b188a9b9d9527ecd70ab4330cf4970b4

okwme:
Made https://github.com/denalimarsh/peggy/pull/3 to hopefully fix issue running tests in CI

denalimarsh:
@okwme looks like this is cleared for merge

PR #182, opened 2020-04-13
Dewhitelisting modules

musnit:
Partially Closes: #123  (Relayer and Solidity code still TODO)

This PR modifies the cosmos modules/sdk-related code to remove support for whitelisted tokens. 
 - Token contract address is no longer part of the burn/lock APIs.
 - Tokens minted when coming from Ethereum through lockups on Ethereum are minted with the 'peggy' prefix
 - Tokens being burned, to be sent back to Ethereum, need to have the 'peggy' prefix. Normal cosmos tokens can no longer be burned
 - CLI/APIs now use separate fields for amount and symbol, rather than a single field

______

For contributor use:

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [x] Wrote unit and integration [tests](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#testing)
- [x] Updated relevant documentation (`docs/`) or specification (`x/<module>/spec/`)
- [x] Added  relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code).
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer

______

For admin use:

- [ ] Added appropriate labels to PR (ex. `WIP`, `R4R`, `docs`, etc)
- [ ] Reviewers assigned
- [ ] Squashed all commits, uses message "Merge pull request #XYZ: [title]" ([coding standards](https://github.com/tendermint/coding/blob/master/README.md#merging-a-pr))

### DIFF
--- a/x/ethbridge/client/cli/tx.go
+++ b/x/ethbridge/client/cli/tx.go
@@ -2,9 +2,12 @@ package cli
 
 import (
 	"bufio"
+	"regexp"
 	"strconv"
 
 	"github.com/cosmos/peggy/x/ethbridge/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
@@ -36,8 +39,14 @@ func GetCmdCreateEthBridgeClaim(cdc *codec.Codec) *cobra.Command {
 			}
 
 			tokenContractString := viper.GetString(types.FlagTokenContractAddr)
+			if !common.IsHexAddress(tokenContractString) {
+				return errors.Errorf("invalid [token-contract-address]: %s", tokenContractString)
+			}
 			tokenContract := types.NewEthereumAddress(tokenContractString)
 
+			if !common.IsHexAddress(args[0]) {
+				return errors.Errorf("invalid [bridge-registry-contract]: %s", args[0])
+			}
 			bridgeContract := types.NewEthereumAddress(args[0])
 
 			nonce, err := strconv.Atoi(args[1])
@@ -47,6 +56,9 @@ func GetCmdCreateEthBridgeClaim(cdc *codec.Codec) *cobra.Command {
 
 			symbol := args[2]
 			ethereumSender := types.NewEthereumAddress(args[3])
+			if !common.IsHexAddress(args[3]) {
+				return errors.Errorf("invalid [ethereum-sender-address]: %s", args[0])
+			}
 			cosmosReceiver, err := sdk.AccAddressFromBech32(args[4])
 			if err != nil {
 				return err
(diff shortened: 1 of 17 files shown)


okwme:
is this R4R? I see some tests failing

musnit:
@okwme yes is. tests that are failing are part of the relayer/contract stuff that will be fixed in a different PR.

x/ethbridge/client/cli/tx.go
```diff
@@ -100,21 +120,31 @@ func GetCmdBurn(cdc *codec.Codec) *cobra.Command {
...
+			}
+			amount, err := strconv.ParseInt(args[2], 10, 64)
 			if err != nil {
 				return err
 			}
+			if amount <= 0 {
+				return types.ErrInvalidAmount
+			}
+
+			symbol := args[3]
```
okwme, line 145: do we need to check if this string is empty?

okwme:
> CLI/APIs now use separate fields for amount and symbol, rather than a single field
What was the thinking behind this change? doesn't the Coin type allow for this already?

musnit:
> > CLI/APIs now use separate fields for amount and symbol, rather than a single field
> 
> What was the thinking behind this change? doesn't the Coin type allow for this already?

Initial thinking was to improve the cli/apis/interface so that it's not "12peg::address::tokenname" and can just be 12 tokenname

x/ethbridge/client/cli/tx.go
musnit: This is effectively checked by the prefix/suffix checks in MsgBurn ValidateBasic, but yeh may b good to check here in CLI too.

okwme:
the code updates look good and the tests all pass. Not sure if we should merge into master until other tests pass though

musnit:
@okwme cool yeh - tests all pass in https://github.com/cosmos/peggy/pull/185 which include these changes.

okwme:
@musnit aha, then I'm confused about the order of operations here. Shouldn't this and the other PR be going into the branch on #185? And then 185 should go into master?

musnit:
@okwme yep that's correct. we can deprecate this PR if we merge #185

okwme:
that sounds good i'll start reviewing it now, think there's also a PR for the contract tests waiting for @denalimarsh

PR #189, opened 2020-04-24
migrate codecov to github actions

marbar3778:
migrate somethings to github actions


marbar3778:
not sure if the codecov was being uploaded before?

marbar3778:
we should add a codecov badge to readme when there is a codecov report for master

PR #230, opened 2020-09-07
UpdateValset, SubmitBatch tests and writeup

jtremback:
This finishes the detailed UpdateValset and SubmitBatch tests, and adds a writeup explaining the contract. It also adds a gas test comparing just saving the entire valset instead of making a checkpoint.


jkilpatr:
rebased and merged

PR #242, opened 2020-10-17
WIP multi-erc20

jtremback:
This WIP does a few things

- Switch to multi-ERC20 support
- Eliminate updateValset and submitBatch in favor of combo method only
- Add a couple if statements to allow for empty batches to be submitted in the combo method

There's some question of whether it's a good idea to add complexity to the combo method to eliminate updateValset. I think it probably is because the complexity is not much. It's just easier and less error prone to maintain one function and tests.


jtremback:
PR'ed to wrong repo lol

PR #250, opened 2020-11-09
update soliditiy docs

jtremback:


jtremback:
lol i keep doing these on the wrong repo. Thanks Github.

PR #266, opened 2021-02-24
V2 minor usability junk

jtremback:
This adds routes to query the ERC20 to Denom mapping, as well as the code to interface it with genesis. 

This stuff is untested- We don't have any unit tests for routes and genesis. Wondering what the usual technique for testing it is. Add it to the integration tests?


jtremback:
Argh wrong repo

PR #278, opened 2021-03-12
rename peggy to gravity

marbar3778:


fedekunze (approved):
utACK

marbar3778:
will clean this up

jkilpatr:
I will be reverting this change it doesn't work at all.

marbar3778:
which part doesnt work? We can spend sometime fixing the tests instead of reverting.

jkilpatr:
Message scopes need to be fixed in the orchestrator (gravity/MsgWhatever versus peggy/MsgWhatever, I don't think there's too much more to do. But we have no idea if there is because the tests are just obviously broken, you can't just merge a patch this big and not even look at the tests.

We have no idea if this works all up and 'merge it and make it work after' is not an acceptable coding practice.

marbar3778:
> Message scopes need to be fixed in the orchestrator (gravity/MsgWhatever versus peggy/MsgWhatever), I don't think there's too much more to do. But we have no idea if there is because the tests are just obviously broken, you can't just merge a patch this big and not even look at the tests.
> 
> We have no idea if this works all up and 'merge it and make it work after' is not an acceptable coding practice.

I'll set some ci restrictions to avoid this in the future. No need to revert.

marbar3778:
> Message scopes need to be fixed in the orchestrator (gravity/MsgWhatever versus peggy/MsgWhatever),

I can't find this any messages with peggy/Msg*. These were changed.

jkilpatr:
> I'll set some ci restrictions to avoid this in the future. No need to revert. No need to bring coding practice. This repo is not setup correctly, and we are all trying our best to instill best practices

The tests still have some flake issues (although those have been greatly reduce recently) and the darn arbitrary logic test needs a repo secret that's not avaialble to outside PRs (so it will always fail if someone external opens a PR).

My opinion on CI design after being a CI engineer for a couple of years is that it must always be up to engineers to look at the output and understand it and the managers to hold their reports responsible for not doing so. Trying to run greenlight / redlight CI systems was just a mess at scale.



> > Message scopes need to be fixed in the orchestrator (gravity/MsgWhatever versus peggy/MsgWhatever),
> 
> I can't find this any messages with peggy/Msg*. These were changed.

I see that now. With the obvious out of the way that means there's some more subtle bug.

marbar3778:
working on that now 👍 Noticed that after trying to run cargo build a few times 😆

marbar3778:
working on it here https://github.com/cosmos/gravity-bridge/pull/306/files

jkilpatr:
I think I've got it. There's on trick to it. The proto regen, which should be in a readme somewhere.

jkilpatr:
there was also no renaming in the solidity directory. so everything over there needs to be done.

jkilpatr:
There we go looks like this is resolved. I'll rebase the rest of the PRs this morning. You did a good job with this pr @marbar3778 it just needed a little more attention to get everything working.

Just pm'ing me that you wanted this merged would have been enough to get my attention and this fixed up.

If we had been unlucky and some unexpected problem was waiting in the wings this could have borked main for days or weeks. Putting us in a bad spot between reverting and losing work and having anything that works at all. Considering we are all actively running testnets this isn't acceptable.

I know things are running behind schedule here and it does take me time to get to PRs and help explain how some of the other Gravity components work. But we're all professionals here, I don't think we need to hide the merge button like we would from a child. It's reasonable to expect everyone to be conscientious and capable.

marbar3778:
> Just pm'ing me that you wanted this merged would have been enough to get my attention and this fixed up.

Sorry about that. Tried to set up the bot to request the needed people, may not have worked correctly. Will ping you in tg next time 😄

PR #283, opened 2021-03-12
proto: regenerate

marbar3778:
## Description

Regenerate proto messages to account for security vulnerability


fedekunze (approved):

jkilpatr:
this branch duplicates the add codeowners commit which was merged in it's own PR. I'll be removing it. otherwise it's just formatting changes and that's fine.

Not approving because it requires changes before merging. I'll perform those changes though.

jkilpatr:
I stand corrected, there's a non-formatting change that I can't really determine the need for.

What is 'buf' doing? Why should we add it to our workflow? You don't seem to have made any actual proto changes so if it's a proto linter it didn't actually catch anything or generate any checked in files (to for example check for not compatible changes).

It seems that this change must be either incomplete, or not really do anything.

jkilpatr (changes requested):
there's a non-formatting change that I can't really determine the need for.

What is 'buf' doing? Why should we add it to our workflow? You don't seem to have made any actual proto changes so if it's a proto linter it didn't actually catch anything or generate any checked in files (to for example check for not compatible changes).

It seems that this change must be either incomplete, or not really do anything.

marbar3778:
this is generating code with gogoproto v1.3.3. I mentioned in the description it is for a secruity vulnerability

```
Regenerate proto messages to account for **security vulnerability**
```

I use buf to avoid the need to install protoc. It does the samething. We should migrate to buf gen entirely to avoid using protoc

jkilpatr:
> this is generating code with gogoproto v1.3.3. I mentioned in the description it is for a secruity vulnerability
> 
> ```
> Regenerate proto messages to account for **security vulnerability**
> ```
> 
> I use buf to avoid the need to install protoc. It does the samething. We should migrate to buf gen entirely to avoid using protoc

If we're just using buff to avoid protoc isn't it strictly inferior? More complexity for the same output since buff contains protoc + more stuff? Seems like dependency creep.

The real problem here is how to keep a dev with an older protoc from reverting these security fixes.

fedekunze:
> there's a non-formatting change that I can't really determine the need for.

The proto format fixes the format with docker and a `clang-format` file so that everyone shares the same format when running the proto generating tools

>If we're just using buff to avoid protoc isn't it strictly inferior? More complexity for the same output since buff contains protoc + more stuff? Seems like dependency creep.
The real problem here is how to keep a dev with an older protoc from reverting these security fixes.

Is someone running the bridge on production? these tools are being used on the latest hub upgrade so we can be sure they are reliable

jkilpatr:
> Is someone running the bridge on production? these tools are being used on the latest hub upgrade so we can be sure they are reliable

This pull request claims to fix a security vulnerability. What it actually does is add a new tool, and band-aid a security vulnerability such that it could come back later. What tooling does the SDK use to make sure no one checks in proto generated with an insecure version of gogoproto?

> This isn't due to protoc but gogoproto. Protoc generates file descriptors and gogoproto generates the go code. There isn't any script that install any proto dependencies other than protoc in this repo. This should be fixed, right now it is unclear which versions of `--grpc-gateway_out` and `--gocosmos_out` to use or how to download them. Would be good to add some scripts or just use the docker file we use in the SDK.

Agreed, `make proto-gen` should fail if the user has an earlier version of protoc than that's used in the repo. We should also have a proper tool for installing/updating utils.

> I wouldn't say buf is dependency creep. We use it throughout the cosmos ecosystem because it forces you to use best practices. Also it's already being used in the repo https://github.com/cosmos/gravity-bridge/blob/main/module/Makefile#L73, so reducing the dependency from buf and protoc to only buf seems like the right approach.

So far the reasons for using buf that have been presented to me are. 'we already use it', 'it does the same thing', 'the ecosystem uses it', and 'it forces you to use best practices'. None of these are valid engineering reasons, no concrete problem has been presented that it solves.

At this point I'm just wondering why we're spending time talking about it if it's so hard to give a concrete example of what it does that's an improvement to the repo.

We have a concrete, known way that a security vulnerability could be easily introduced into a production gravity bridge, our proto generation tooling needs to be improved and check for more things. Seems much more productive to focus our efforts there.

marbar3778:
Ill revert the buf addition. The security vulnerability was by gogoproto, literally everyone who uses gogoproto updated, and you are trying to avoid updating. https://github.com/gogo/protobuf/commit/b03c65ea87cdc3521ede29f62fe3ce239267c1bc. I'm not sure why I have to convince you to merge an update to a dependency, this is very bad practice IMO.

jackzampolin:
@jkilpatr we try to all use the same proto tooling as an ecosystem. What @marbar3778 and @fedekunze are suggesting is best practice from other projects that we also rely on and forking tooling setups will lead to additional maintenance. Why can't we just go with what Marko is suggesting?

marbar3778:
closing this PR. I will open a PR that sets up the repo for success with proto.

jkilpatr:
I'm not opposed to buff. Or even the entire contents of this PR. Clearly we need to get on a call and hash this out.

PR #287, opened 2021-03-15
client: refactor client

marbar3778:
## Description

Refactor to use any for claims and confirms.


jackzampolin:
@alexanderbez

module/proto/peggy/v1/claims.proto
```diff
@@ -0,0 +1,66 @@
...
+  string token_contract = 4;
+  string orchestrator   = 5;
+}
+
+// This informs the Cosmos module that a logic
+// call has been executed
+message LogicCallExecutedClaim {
+  uint64 event_nonce        = 1;
+  uint64 block_height       = 2;
+  bytes  invalidation_id    = 3;
```
alexanderbez, line 49: why is an ID bytes?

alexanderbez (commented):
I only looked at the proto files thus far and those LGTM 👍

module/proto/peggy/v1/claims.proto
fedekunze: right, it should be an integer

module/proto/peggy/v1/claims.proto
```diff
@@ -0,0 +1,66 @@
...
+  ERC20_DEPLOYED      = 3;
+  LOGIC_CALL_EXECUTED = 4;
+}
+
+// EthereumBridgeDepositClaim
+// When more than 66% of the active validator set has
+// claimed to have seen the deposit enter the ethereum blockchain coins are
+// issued to the Cosmos address in question
+// -------------
+message DepositClaim {
```
fedekunze, line 21: ideally, all the fields should have a comment for reference

module/proto/peggy/v1/confirm.proto
```diff
@@ -0,0 +1,66 @@
...
+// cutting off when the batch either reaches a hardcoded maximum size (to be
+// decided, probably around 100) or when transactions stop being profitable
+// (TODO determine this without nondeterminism) This message includes the batch
+// as well as an Ethereum signature over this batch by the validator
+// -------------
+message ConfirmLogicCall {
+  string invalidation_id    = 1;
+  uint64 invalidation_nonce = 2;
+  string eth_signer         = 3;
+  string orchestrator       = 4;
```
fedekunze, line 26: nit, can we rename these to `eth_address` and `orchestrator_address`?

module/proto/peggy/v1/confirm.proto
```diff
@@ -0,0 +1,66 @@
...
+  uint64 nonce          = 1;
+  string token_contract = 2;
+  string eth_signer     = 3;
+  string orchestrator   = 4;
+  string signature      = 5;
+}
+
+// ValsetConfirm
+// this is the message sent by the validators when they wish to submit their
+// signatures over the validator set at a given block height. A validator must
```
fedekunze, line 48: validators on cosmos? I don't understand the "submit signatures over the validator set". Why do you need to request and confirm the validator set in the first place?

module/proto/peggy/v1/genesis.proto
```diff
@@ -70,51 +71,49 @@ option go_package = "github.com/althea-net/peggy/module/x/peggy/types";
...
+  Params                      params              = 1;
+  uint64                      last_observed_nonce = 2;
+  repeated Valset             valsets             = 3;
+  repeated ValsetConfirm      valset_confirms     = 4;
+  repeated OutgoingTxBatch    batches             = 5;
+  repeated ConfirmBatch       batch_confirms      = 6 [(gogoproto.nullable) = false];
+  repeated OutgoingLogicCall  logic_calls         = 7;
+  repeated ConfirmLogicCall   logic_call_confirms = 8 [(gogoproto.nullable) = false];
+  repeated Attestation        attestations        = 9 [(gogoproto.nullable) = false];
+  repeated MsgDelegateKey     delegate_keys       = 10;
```
fedekunze, line 116: we shouldn't use msgs here

module/proto/peggy/v1/genesis.proto
```diff
@@ -70,51 +71,49 @@ option go_package = "github.com/althea-net/peggy/module/x/peggy/types";
...
-  repeated OutgoingTransferTx        unbatched_transfers = 12;
+  Params                      params              = 1;
+  uint64                      last_observed_nonce = 2;
+  repeated Valset             valsets             = 3;
+  repeated ValsetConfirm      valset_confirms     = 4;
+  repeated OutgoingTxBatch    batches             = 5;
+  repeated ConfirmBatch       batch_confirms      = 6 [(gogoproto.nullable) = false];
+  repeated OutgoingLogicCall  logic_calls         = 7;
+  repeated ConfirmLogicCall   logic_call_confirms = 8 [(gogoproto.nullable) = false];
+  repeated Attestation        attestations        = 9 [(gogoproto.nullable) = false];
```
fedekunze, line 115: why aren't we using Any here instead of these individual types?

module/proto/peggy/v1/genesis.proto
```diff
@@ -70,51 +71,49 @@ option go_package = "github.com/althea-net/peggy/module/x/peggy/types";
...
+  repeated MsgDelegateKey     delegate_keys       = 10;
+  repeated ERC20ToDenom       erc20_to_denoms     = 11;
```
fedekunze, line 117: I don't understand what this is for

module/proto/peggy/v1/genesis.proto
```diff
@@ -70,51 +71,49 @@ option go_package = "github.com/althea-net/peggy/module/x/peggy/types";
...
-  string bridge_ethereum_address          = 4;
-  uint64 bridge_chain_id                  = 5;
-  uint64 signed_valsets_window            = 6;
-  uint64 signed_batches_window            = 7;
-  uint64 signed_claims_window             = 8;
-  uint64 target_batch_timeout             = 10;
-  uint64 average_block_time               = 11;
-  uint64 average_ethereum_block_time      = 12;
-  bytes  slash_fraction_valset            = 13  [
+  string peggy_id                    = 1;
```
fedekunze, line 76:
```suggestion
  string gravity_id                    = 1;
```

module/proto/peggy/v1/genesis.proto
```diff
@@ -70,51 +71,49 @@ option go_package = "github.com/althea-net/peggy/module/x/peggy/types";
...
+  string contract_source_hash        = 2;
+  string bridge_ethereum_address     = 4;
+  uint64 bridge_chain_id             = 5;
```
fedekunze, line 79:
```suggestion
  uint64 eip155_chain_id             = 5;
```

module/proto/peggy/v1/pool.proto
```diff
@@ -6,9 +6,11 @@ import "gogoproto/gogo.proto";
...
+}
 
-message BatchFees {  
-  string token = 1;  
+message BatchFees {
+  string token         = 1;
   string topOneHundred = 2 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int", (gogoproto.nullable) = false];
-}
\ No newline at end of file
+}
```
fedekunze: what are these for?

module/x/peggy/types/claims.go
```diff
@@ -0,0 +1,197 @@
...
+func (b *WithdrawClaim) ClaimHash() []byte {
+	path := fmt.Sprintf("%s/%d/", b.TokenContract, b.BatchNonce)
+	return tmhash.Sum([]byte(path))
+}
+
+func (msg WithdrawClaim) GetClaimer() sdk.AccAddress {
+	err := msg.ValidateBasic()
+	if err != nil {
+		panic("WithdrawClaim failed ValidateBasic! Should have been handled earlier")
+	}
```
fedekunze, line 117: imho this shouldn't be part of this function

module/x/peggy/types/claims.go
```diff
@@ -0,0 +1,197 @@
...
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, e.Orchestrator)
+	}
+	if e.EventNonce == 0 {
+		return fmt.Errorf("nonce == 0")
+	}
+	return nil
+}
+
+func (msg ERC20DeployedClaim) GetClaimer() sdk.AccAddress {
+	err := msg.ValidateBasic()
```
fedekunze, line 149: ditto

module/x/peggy/types/claims.go
```diff
@@ -0,0 +1,197 @@
...
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, e.Orchestrator)
+	}
+	if e.EventNonce == 0 {
+		return fmt.Errorf("nonce == 0")
+	}
+	return nil
+}
+
+func (msg LogicCallExecutedClaim) GetClaimer() sdk.AccAddress {
+	err := msg.ValidateBasic()
```
fedekunze, line 184: ditto

module/x/peggy/types/confirm.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return sdkerrors.Wrapf(sdkerrors.ErrUnknownRequest, "Could not decode hex string %s", msg.Signature)
+	}
+	return nil
+}
+
+// GetSignBytes encodes the message for signing
+func (msg ConfirmBatch) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(msg))
+}
```
fedekunze, line 45: ?? why are we using amino?

module/x/peggy/types/confirm.go
```diff
@@ -0,0 +1,135 @@
...
+
+// GetSignBytes encodes the message for signing
+func (msg *ValsetConfirm) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(msg))
+}
+
+// GetSigners defines whose signature is required
+func (msg *ValsetConfirm) GetSigners() []sdk.AccAddress {
+	// TODO: figure out how to convert between AccAddress and ValAddress properly
```
fedekunze, line 129: you can just use `sdk.ValAddress(accAddr)`

module/x/peggy/types/msgs.go
```diff
@@ -199,387 +152,166 @@ func (msg MsgRequestBatch) GetSigners() []sdk.AccAddress {
...
-		return sdkerrors.Wrap(err, "erc20 token")
-	}
-	if _, err := sdk.AccAddressFromBech32(e.Orchestrator); err != nil {
-		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, e.Orchestrator)
+func (msg *MsgSubmitConfirm) ValidateBasic() (err error) {
+	_, err = sdk.ValAddressFromBech32(msg.Signer)
+	if err != nil {
+		return err
 	}
 	return nil
```
fedekunze, line 211: you should validate the confirm field

module/x/peggy/types/msgs.go
```diff
@@ -199,387 +152,166 @@ func (msg MsgRequestBatch) GetSigners() []sdk.AccAddress {
...
-	if e.BatchNonce == 0 {
-		return fmt.Errorf("batch_nonce == 0")
-	}
-	if err := ValidateEthAddress(e.TokenContract); err != nil {
-		return sdkerrors.Wrap(err, "erc20 token")
-	}
-	if _, err := sdk.AccAddressFromBech32(e.Orchestrator); err != nil {
-		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, e.Orchestrator)
+func (msg *MsgSubmitConfirm) ValidateBasic() (err error) {
+	_, err = sdk.ValAddressFromBech32(msg.Signer)
```
fedekunze, line 207: is val address here ok?

module/x/peggy/types/msgs.go
```diff
@@ -199,387 +152,166 @@ func (msg MsgRequestBatch) GetSigners() []sdk.AccAddress {
...
-		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, e.Orchestrator)
-	}
-	if e.EventNonce == 0 {
-		return fmt.Errorf("nonce == 0")
+func (msg *MsgSubmitClaim) ValidateBasic() (err error) {
+	_, err = sdk.ValAddressFromBech32(msg.Signer)
+	if err != nil {
+		return err
 	}
 	return nil
```
fedekunze, line 275: validate claim

module/x/peggy/types/confirm.go
marbar3778: No need to review the code yet, it still needs to be properly cleaned up

module/proto/peggy/v1/claims.proto
marbar3778: current gravity design.

module/proto/peggy/v1/claims.proto
jackzampolin: we can add this later

module/proto/peggy/v1/claims.proto
jackzampolin: uint64 ++

module/proto/peggy/v1/confirm.proto
jackzampolin: agreed

module/proto/peggy/v1/confirm.proto
jackzampolin: This is because we need to keep the owners of the contract on ETH the same as the valset on cosmos. Each cosmos validator has an eth key associated and that makes up the multisig

module/proto/peggy/v1/genesis.proto
jackzampolin: yasss rename !~~~

module/proto/peggy/v1/genesis.proto
jackzampolin: Attestation is a wrapper type that includes the claims afaik

module/proto/peggy/v1/genesis.proto
jackzampolin: yasss rename !~~~

module/proto/peggy/v1/genesis.proto
jackzampolin: why is this bad? if the data structure works can't we just reuse code?

module/proto/peggy/v1/genesis.proto
jackzampolin: this has to do with the cosmos > eth transfers where we create an ERC20 on the ETH side and send cosmos denoms over to eth (i.e. ATOMs on Uniswap).

module/x/peggy/types/confirm.go
jackzampolin: 🔪  AMINO 🔪 

proto

jackzampolin (approved):
This gets an ACK from me!

module/x/peggy/keeper/claim_handler.go
```diff
@@ -0,0 +1,215 @@
...
+	any, err := codectypes.NewAnyWithValue(msg)
+	if err != nil {
+		return err
+	}
+
+	// Add the claim to the store
+	_, err = k.Attest(ctx, claim, any)
+	if err != nil {
+		return sdkerrors.Wrap(err, "create attestation")
+	}
```
marbar3778, line 52: @fedekunze can you check this is correct. The any stuff is a bit weird.

module/proto/peggy/v1/confirm.proto
fedekunze: I understand that but why don't we submit a single `UpdateValset` tx to ethereum with the keys and updated power instead? The request + confirm design is something that I really don't get

module/proto/peggy/v1/genesis.proto
fedekunze: I see, what about the confirms?

marbar3778: I would prefer to keep the scope of internal changes in this PR to a minimum as it will become hard to review with just the API changes. Please open an issue and let's add it to the project board.

module/contrib/local/protocgen.sh
```diff
@@ -9,15 +9,9 @@ for dir in $proto_dirs; do
   -I "third_party/proto" \
   --gocosmos_out=plugins=interfacetype+grpc,\
 Mgoogle/protobuf/any.proto=github.com/cosmos/cosmos-sdk/codec/types:. \
+  --grpc-gateway_out=logtostderr=true:. \
   $(find "${dir}" -maxdepth 1 -name '*.proto')
-
-  # # command to generate gRPC gateway (*.pb.gw.go in respective modules) files
-  # buf protoc \
```
fedekunze, line 15: Does this work?

module/proto/gravity/v1/attestation.proto
```diff
@@ -39,7 +29,9 @@ message Attestation {
   bool                observed = 1;
   repeated string     votes    = 2;
   uint64              height   = 3;
-  google.protobuf.Any claim    = 4;
+  google.protobuf.Any claim    = 4 [
+    (cosmos_proto.accepts_interface) = "EthereumClaim"
```
fedekunze, line 33: I haven't seen this option work

module/x/gravity/keeper/claim_handler.go
```diff
@@ -0,0 +1,215 @@
...
+	// return an error if the validator isn't in the active set
+	val := k.StakingKeeper.Validator(ctx, validator)
+	if val == nil || !val.IsBonded() {
+		return sdkerrors.Wrap(sdkerrors.ErrorInvalidSigner, "validator not in acitve set")
+	}
+
+	msg, ok := claim.(proto.Message)
+	if !ok {
+		return fmt.Errorf("can't proto marshal %T", msg)
+	}
```
fedekunze, line 93: you don't need to add this if you embed `proto.Message` to the interface

module/x/gravity/keeper/claim_handler.go
```diff
@@ -0,0 +1,215 @@
...
+		return sdkerrors.Wrap(err, "create attestation")
+	}
+
+	// Emit the handle message event
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, claim.GetType().String()),
+			// TODO: maybe return something better here? is this the right string representation?
+			sdk.NewAttribute(types.AttributeKeyAttestationID, string(types.GetAttestationKey(claim.GetEventNonce(), claim.ClaimHash()))),
```
fedekunze, line 112: I really don't understand this event value. Why isn't this split into 2 attr?

module/x/gravity/keeper/claim_handler.go
```diff
@@ -0,0 +1,215 @@
...
+	// return an error if the validator isn't in the active set
+	val := k.StakingKeeper.Validator(ctx, validator)
+	if val == nil || !val.IsBonded() {
+		return sdkerrors.Wrap(sdkerrors.ErrorInvalidSigner, "validator not in acitve set")
+	}
+
+	msg, ok := claim.(proto.Message)
+	if !ok {
+		return fmt.Errorf("can't proto marshal %T", msg)
+	}
```
fedekunze, line 142: ditto

module/x/gravity/keeper/claim_handler.go
```diff
@@ -0,0 +1,215 @@
...
+	any, err := codectypes.NewAnyWithValue(msg)
+	if err != nil {
+		return err
+	}
+
+	// Add the claim to the store
+	_, err = k.Attest(ctx, claim, any)
+	if err != nil {
+		return sdkerrors.Wrap(err, "create attestation")
+	}
```
fedekunze, line 153: seems that these funcs have a lot of code duplication

module/x/gravity/keeper/confirm_handler.go
```diff
@@ -0,0 +1,202 @@
...
+	validator := k.GetOrchestratorValidator(ctx, valaddr)
+	if validator == nil {
+		sval := k.StakingKeeper.Validator(ctx, sdk.ValAddress(valaddr))
+		if sval == nil {
+			return sdkerrors.Wrap(types.ErrUnknown, "validator")
+		}
+		validator = sval.GetOperator()
+	}
+
+	ethAddress := k.GetEthAddress(ctx, validator)
```
fedekunze, line 45: imho this should return the `common.Address` struct, but maybe for another PR

module/x/gravity/keeper/confirm_handler.go
```diff
@@ -0,0 +1,202 @@
...
+	batch := k.GetOutgoingTXBatch(ctx, confirm.GetTokenContract(), confirm.GetNonce())
+	if batch == nil {
+		return sdkerrors.Wrap(types.ErrInvalid, "couldn't find batch")
+	}
+
+	peggyID := k.GetGravityID(ctx)
+	checkpoint, err := batch.GetCheckpoint(peggyID)
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrInvalid, "checkpoint generation")
+	}
```
fedekunze, line 28: we should move this below sig verification

module/x/gravity/keeper/confirm_handler.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// check if we already have this confirm
+	if k.GetBatchConfirm(ctx, confirm.GetNonce(), confirm.GetTokenContract(), valaddr) != nil {
+		return sdkerrors.Wrap(types.ErrDuplicate, "duplicate signature")
+	}
+
+	msg, ok := confirm.(*types.ConfirmBatch)
+	if !ok {
+		return fmt.Errorf("can't proto marshal %T", msg)
+	}
```
fedekunze, line 63: isn't this checked already?

module/x/gravity/keeper/confirm_handler.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrInvalid, fmt.Sprintf("signature verification failed expected sig by %s with peggy-id %s with checkpoint %s found %s", ethAddress, peggyID, hex.EncodeToString(checkpoint), confirm.GetSignature()))
+	}
+
+	// check if we already have this confirm
+	if k.GetLogicCallConfirm(ctx, invalidationIdBytes, confirm.GetInvalidationNonce(), valaddr) != nil {
+		return sdkerrors.Wrap(types.ErrDuplicate, "duplicate signature")
+	}
+
+	msg, ok := confirm.(*types.ConfirmLogicCall)
```
fedekunze, line 128: ditto

module/x/gravity/keeper/confirm_handler.go
```diff
@@ -0,0 +1,202 @@
...
+	logic := k.GetOutgoingLogicCall(ctx, invalidationIdBytes, confirm.GetInvalidationNonce())
+	if logic == nil {
+		return sdkerrors.Wrap(types.ErrInvalid, "couldn't find logic")
+	}
+
+	peggyID := k.GetGravityID(ctx)
+	checkpoint, err := logic.GetCheckpoint(peggyID)
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrInvalid, "checkpoint generation")
+	}
```
fedekunze, line 96: ditto move below

module/x/gravity/types/claims.go
```diff
@@ -0,0 +1,199 @@
...
+import (
+	fmt "fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/tendermint/tendermint/crypto/tmhash"
+)
+
+// EthereumClaim represents a claim on ethereum state
+type EthereumClaim interface {
```
fedekunze, line 12: add proto msg interface

fedekunze: add ValidateBasic

module/x/gravity/types/confirm.go
```diff
@@ -0,0 +1,116 @@
+package types
+
+import (
+	"encoding/hex"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+type Confirm interface {
```
fedekunze, line 10: ditto proto and validate basic

module/contrib/local/protocgen.sh
marbar3778: yea, you can add as many generators to a single protoc command as you want

module/proto/peggy/v1/genesis.proto
marbar3778: is this specific to eip155? eip155 was deprecated in geth 1.10, is this a problem?

module/proto/gravity/v1/attestation.proto
marbar3778: This is how the sdk handles any with gov and keys.

module/x/gravity/keeper/claim_handler.go
marbar3778: carry over from before the refactor.

module/proto/gravity/v1/attestation.proto
jackzampolin: I've seen it working fwiw

module/proto/peggy/v1/claims.proto
jkilpatr: This is a design decision by @jtremback think of it as a hash value separating different arbitrary logic execution flows.

It can be a number if it really needs to be IMO but Jehan should weigh in.

module/proto/peggy/v1/confirm.proto
jkilpatr: Requests are coordinating points for the external signers to work with.

If all the orchestrators tried to sign the 'current validator set' they would get different values since they are operating outside of the tendermint consensus code. For validator sets we've automated validator set request creation, but we can't remove it without moving the Ethereum signing into tendermint iself. Which maybe possible in the future from what I've heard.

If we could import an ethereum private key from the validators key store and perform signatures in the module then requests could be eliminated as a concept. I think this should be a long term goal.

For now we're stuck with them.

module/proto/gravity/v1/claims.proto
```diff
@@ -0,0 +1,66 @@
+syntax = "proto3";
```
jkilpatr, line 1: 👍🏻  this is good.

jkilpatr (approved):
Looks good to me. Eager to get this merged and have these interfaces cleaned up.

This should be unblocked once #321 makes it in  🎆

marbar3778:
Tests are passing, need to do one final sweep, then we can merge this into refactor. Will jump in on the orchestrator to get tx submission in.

module/x/peggy/types/confirm.go
marbar3778: amino is now removed.

fedekunze (approved):

PR #290, opened 2021-03-16
Update Geth and enable Berlin

jkilpatr:
Updates Geth to the latest version and enables the Berlin hardfork
on our test chain.


marbar3778 (approved):
LGTM

PR #291, opened 2021-03-17
Dockerizing tests

mvid:


jackzampolin:
@alexanderbez

alexanderbez:
@mvid mind giving a quick instruction set on how to test this? I can obviously start the Docker network via docker-compose, but how then do I test completeness? Are there commands or tests to run? Or is just starting the docker compose network sufficient?

mvid:
@alexanderbez I am still working on completeness for the network, but right now I have a script in the root called `./start-testnet.sh` which should handle genesis, key signing, etc. After that, there will (probably) be another container that runs tests. Or maybe just a script.

alexanderbez:
I see, so if I execute `./start-testnet.sh` with a zero exit-code, I can deem this PR acceptable?

ethereum/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+           --mine \
+           --http \
+           --http.port "8545" \
+           --http.addr "0.0.0.0" \
+           --http.corsdomain "*" \
+           --http.vhosts "*" \
+           --miner.threads=1 \
+           --nousb \
+           --verbosity=5 \
+           --miner.etherbase=0xBf660843528035a5A4921534E156a27e64B231fE
```
jackzampolin, line 22: This is a fun magic number.

solidity/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM node:15.11-alpine3.13
+
+RUN apk update
+RUN apk add --no-cache python3 make g++ curl
+
+COPY . .
+RUN npm ci
+RUN chmod -R +x scripts
+
+RUN npm run typechain
```
jackzampolin, line 10: Is there a way to do `npm` dependency caching here?

jackzampolin: Related, has this been an issue as far as timing?

start-testnet.sh
```diff
@@ -0,0 +1,227 @@
...
+echo "$n2name"_COSMOS_RPC="http://$n2name:26657" >> $n2dir/orchestrator.env
+echo "$n2name"_COSMOS_KEY=$(jq .priv_key.value $n2cfgDir/priv_validator_key.json) >> $n2dir/orchestrator.env
+echo "$n2name"_DENOM=stake >> $n2dir/orchestrator.env
+echo "$n2name"_ETH_RPC=http://ethereum:8545 >> $n2dir/orchestrator.env
+
+echo "$n3name"_COSMOS_GRPC="http://$n3name:9090" >> $n3dir/orchestrator.env
+echo "$n3name"_COSMOS_RPC="http://$n3name:26657" >> $n3dir/orchestrator.env
+echo "$n3name"_COSMOS_KEY=$(jq .priv_key.value $n3cfgDir/priv_validator_key.json) >> $n3dir/orchestrator.env
+echo "$n3name"_DENOM=stake >> $n3dir/orchestrator.env
+echo "$n3name"_ETH_RPC=http://ethereum:8545 >> $n3dir/orchestrator.env
```
jackzampolin, line 207: Might be nice to break the above into a separate script where you could pass in a directory to populate as an arg.

solidity/scripts/contract-deployer.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 npx ts-node \
 contract-deployer.ts \
---cosmos-node="http://localhost:26657" \
+--cosmos-node="http:///localhost:26657" \
```
jackzampolin, line 4: ?

ethereum/Dockerfile
mvid: I got it from here: https://github.com/cosmos/gravity-bridge/blob/fd98d205b8db4575112b29735d77ecd569d67a1b/solidity/scripts/contract-deployer.sh#L6

If it doesn't have purpose, I will remove it. But it was there before

solidity/Dockerfile
mvid: It hasn't been an issue as far as timing. Also it should be cached by the docker build if nothing before it changed

start-testnet.sh
mvid: Yes, we should do that, and general prettyfication, after it works

solidity/scripts/contract-deployer.sh
mvid: No idea. Removed.

solidity/Dockerfile
mvid: So far the slowest parts are deploying the contracts and building the orchestrator (rust build)

zmanian:
To run `./start-testnet.sh` on my Mac I needed to 

`cd module && make install`
`brew install sponge`

`docker-compose logs orchestrator0` shows the timeout of the orachestor trying to communicate with `http://gravity0:1317`

A debug message shows the url is correct up to being passed to the Contact library

```
orchestrator0_1      | [peggy_utils/src/connection_prep.rs:109] &cosmos_legacy_url = "http://gravity0:1317"
```

zmanian:
This is the only thing I've found in the dependencies for the orchestrator that might be an explanation 

https://github.com/actix/actix-web/issues/599

zmanian:
@jkilpatr could you take a look? Does this meet your testing needs?

zmanian:
@mvid can you fix the conflicts?

mvid:
@zmanian conflicts fixed

mvid:
Seeing this error in all but the first orchestrator:
`Valset cost estimate for Nonce 1 failed with Err(EthereumRestError(JsonRPCError { code: 3, message: "execution reverted: Supplied current validators and powers do not match checkpoint`

Any ideas @jkilpatr ?

jackzampolin:
This looks like it is passing checks. Going to merge into the gentx branch

PR #299, opened 2021-03-22
ci: separate build and test

marbar3778:
## Description

Currently, Build and Test are the same job. This PR separates the two in order to help reviewers hone into what is failing quicker. 

- only run rust tests and build when rust files are touched. This helps with faster build pipelines with other jobs that are more relevant when working in go


.github/workflows/rust.yml
```diff
@@ -2,9 +2,11 @@ name: Rust
 
 on:
   push:
-    branches: [master, main]
+    branches:
+      - main
   pull_request:
-    branches: [master, main]
+    paths:
```
marbar3778, line 8: @jkilpatr is it fine to only run rust tests when the files are touched? Want to make sure these tests are dependent on things in the module directory.

fedekunze (approved):

PR #300, opened 2021-03-22
proto: add buf.yaml, fix linting

marbar3778:
## Description

- Add buf.yaml with rules for linting
- Fix make proto-lint. Avoid needing to download buf. 
- Lint proto files
- Add protobuf linting and break checking to ci

### DIFF
--- a/module/proto/peggy/v1/msgs.proto
+++ b/module/proto/peggy/v1/msgs.proto
@@ -5,7 +5,7 @@ import "gogoproto/gogo.proto";
 import "google/api/annotations.proto";
 option go_package = "github.com/althea-net/peggy/module/x/peggy/types";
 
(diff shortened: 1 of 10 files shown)


module/proto/peggy/v1/msgs.proto
```diff
@@ -5,6 +5,7 @@ import "gogoproto/gogo.proto";
 import "google/api/annotations.proto";
 option go_package = "github.com/althea-net/peggy/module/x/peggy/types";
 
+// Msh defines the state transitions possible within gravity
```
fedekunze, line 8:
```suggestion
// Msg defines the state transitions possible within gravity
```

fedekunze (approved):

PR #303, opened 2021-03-23
Always create more profitable batches

jkilpatr:
This is an important design criteria for permission-less batch creation.

Suppose we have an infinite stream of low fee spam transactions going
into the Gravity tx pool. Any given batch creation event will then lock
up high fee 'good transactions' with dozens of spam transactions.

In this case no relayer will ever chose to relay a batch because all of
them will be unprofitable.

In order to ensure that profitable batches are eventually created we
must avoid locking up the high fee 'good transactions' into obviously
bad batches. To add to the difficulty we don't actually know what any
token in this process is worth or what ETH gas costs.

The solution this patch provides is to ensure that a new batch always
has higher fees than the last batch. This means that even if there are
infinite spam transactions, and infinite spamming of the permission less
create batch request batch creation will halt long enough for enough
good transactions to build up in the pool and a new more profitable
batch to be created.

Over a long enough timescale the old batches either time out. Returning
profitable transactions to the pool to create a new even better batch. Or
the profitability of the latest batch continues to increase until one is
relayed, also freeing the previous good transactions to go into the next
batch.

So given this condition we can say that no matter the inputs a
successful batch will eventually be created.


marbar3778 (changes requested):
Please add a doc explaining the design. We need to get the design written down to allow more people to onboard into this repo.

jkilpatr:
> Please add a doc explaining the design. We need to get the design written down to allow more people to onboard into this repo.

I outline most of how the design works in commit messages. Github doesn't really display commit messages prominently but I'll just copy it out.

marbar3778:
> > Please add a doc explaining the design. We need to get the design written down to allow more people to onboard into this repo.
> 
> I outline most of how the design works in commit messages. Github doesn't really display commit messages prominently but I'll just copy it out.

Easier to read a markdown than 5 git messages.

marbar3778 (approved):
nice addition!! excited to test this with a testnet

PR #304, opened 2021-03-24
Module Import Paths

jackzampolin:
This PR reconfigures the module import paths to point to this repository. This is required to import it from external repositories. This will affect @jkilpatr and @jtremback who are merging into this repo from a fork. If there is a common strategy for dealing this I have not seen it. cc @marbar3778 or @fedekunze maybe?


fedekunze (approved):
ACK, although you might want to update the proto files too

PR #302, opened 2021-03-23
Custom Gentx

jackzampolin:
This PR changes the gentx functionality to require that validators have their ETH key at network startup.

PR #308, opened 2021-03-25
Custom Gentx

jkilpatr:
#302  but rebased and with test fixes.


marbar3778 (approved):
utACK. we should be good to merge this right?

jackzampolin:
We should. @mvid has it working in his docker refactor branch.

jackzampolin (approved):

mvid (approved):

jkilpatr:
This lgtm and I'd like to get it merged but I think we need @fedekunze  to +1 it before it can go in.

marbar3778:
> This lgtm and I'd like to get it merged but I think we need @fedekunze to +1 it before it can go in.

I think someone messed with settings. The codeowners doesn't mean we need to wait for them.

fedekunze (approved):

jkilpatr:
> > This lgtm and I'd like to get it merged but I think we need @fedekunze to +1 it before it can go in.
> 
> I think someone messed with settings. The codeowners doesn't mean we need to wait for them.

I think that's how this is being interpreted.

> Require review from Code Owners Require an approved review in pull requests including files with a designated code owner. 

I think this option has been enabled since we enabled reviews in general. But I'm not totally sure. If the message github is giving us is correct we either need to disable this option or have @jtremback review.

jtremback (commented):
LGTM

PR #309, opened 2021-03-25
proto: use docker 

marbar3778:
## Description

Use docker for proto generation. This helps in avoiding developers installing countless dependencies.


jkilpatr (approved):
looks good 👍🏻  

should resolve issues with older protoc versions being used. Just have to keep an eye on the container for remaining up to date.

PR #317, opened 2021-04-05
module: remove amino

marbar3778:
## Description

Remove amino, this in preparation of proto txs being supported

this is going straight into a feature branch, it will not effect main.


module/x/gravity/types/genesis.go
```diff
@@ -189,9 +188,7 @@ func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 
 // Equal returns a boolean determining if two Params types are identical.
 func (p Params) Equal(p2 Params) bool {
-	bz1 := ModuleCdc.MustMarshalBinaryLengthPrefixed(&p)
-	bz2 := ModuleCdc.MustMarshalBinaryLengthPrefixed(&p2)
-	return bytes.Equal(bz1, bz2)
+	return p.Equal(p2)
```
fedekunze, line 191: this is redundant

fedekunze (approved):
ACK. One comment only

module/x/gravity/types/genesis.go
marbar3778: agree, will fix

PR #319, opened 2021-04-07
fedekunze/state

fedekunze:
- define api
- define claims and confirms
- interfaces
- regenerate proto stubs
- define claim handler, fix cli/tx.go
- claims
- debug tests
- fix error and confirm submisison
- fix naming
- x/gravity: state refactor
- new state (wip)


module/x/gravity/keeper/msg_server.go
```diff
@@ -11,31 +11,16 @@ import (
 	"github.com/cosmos/gravity-bridge/module/x/gravity/types"
 )
 
-type msgServer struct {
```
marbar3778, line 14: Dont we want to keep this to avoid overlap of method names on the keeper itself?

fedekunze: there's no overlap :)

jackzampolin (commented):
Posting some initial read through comments

module/proto/gravity/v1/gravity.proto
```diff
@@ -0,0 +1,151 @@
...
+  repeated EthSigner signers = 1 [(gogoproto.nullable) = false];
+  // TODO: which height? cosmos? This should be the key
+  uint64 height = 2;
+}
+
+// BatchTx represents a batch of transactions going from Cosmos to Ethereum
+message BatchTx {
+  uint64         nonce        = 1;
+  uint64         timeout      = 2;
+  repeated bytes transactions = 3
```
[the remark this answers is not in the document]
fedekunze: the transfer transactions that are included in the batch

module/proto/gravity/v1/confirm.proto
```diff
@@ -1,66 +1,41 @@
...
-// transactions currently in the txqueue in order of highest to lowest fee,
-// cutting off when the batch either reaches a hardcoded maximum size (to be
-// decided, probably around 100) or when transactions stop being profitable
-// (TODO determine this without nondeterminism) This message includes the batch
-// as well as an Ethereum signature over this batch by the validator
-// -------------
+// ConfirmLogicCall ...
 message ConfirmLogicCall {
-  string invalidation_id      = 1;
+  bytes  invalidation_id      = 1 [(gogoproto.casttype) = "github.com/tendermint/tendermint/libs/bytes.HexBytes"];
```
marbar3778, line 10: why use hexbytes here?

fedekunze: because there weren't any docs on why the invalidation id was a plain string so I switched it for consistency

marbar3778: But why the cast type part? this doesnt need to be hexbytes, we use proto json encoding anyways

module/proto/gravity/v1/tx.proto
```diff
@@ -0,0 +1,123 @@
...
+  }
+  rpc SetDelegateKey(MsgDelegateKey) returns (MsgDelegateKeyResponse) {
+    option (google.api.http).post = "/peggy/v1/delegate_keys";
+  }
+}
+
+// MsgTransfer submits a transfer attempt to bridge an asset over to Ethereum.
+// The transfer will be stored and then included in a batch and then
+// submitted to Ethereum.
+message MsgTransfer {
```
marbar3778, line 39: nonce needed?

module/x/gravity/client/cli/tx.go
```diff
@@ -27,74 +24,17 @@ func GetTxCmd(storeKey string) *cobra.Command {
...
-			return nil
-		},
-	}
-}
-
-func CmdSendToEth() *cobra.Command {
+func CmdTransfer() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "send-to-eth [eth-dest] [amount] [bridge-fee]",
+		Use:   "transfer [eth-recipient] [amount] [bridge-fee]",
```
marbar3778, line 37: transfer-to-eth seems to explain what this actually does. Also begs the question, should we support commands that are not needed by validators?

fedekunze: right, I think deleting them is ok

module/proto/gravity/v1/tx.proto
fedekunze: nonce is fetched from the store

module/proto/gravity/v1/confirm.proto
fedekunze: because it casts directly to hex bytes and it's easier to work with them. that's the only reason tbh

jackzampolin: I think this hurts compatibility with other languages so lets try to use the standard types?

module/x/gravity/client/cli/tx.go
jackzampolin: Having a CLI for send to ETH I think is minimum

module/proto/gravity/v1/gravity.proto
```diff
@@ -0,0 +1,151 @@
...
+    (gogoproto.casttype)   = "github.com/tendermint/tendermint/libs/bytes.HexBytes"
+  ];
+  // set of the validator operators address in bech32 format that attest in
+  // favor of this event.
+  repeated string votes = 2;
+  // amount of voting power in support of this event execution
+  int64 attested_power = 3;
+  // height at which the event was attested an executed
+  // TODO: double-check
+  uint64 height = 4;
```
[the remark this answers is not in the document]
jkilpatr: cosmos height.

module/proto/gravity/v1/gravity.proto
```diff
@@ -0,0 +1,151 @@
...
+  bytes tx_id = 1 [(gogoproto.casttype) = "github.com/tendermint/tendermint/libs/bytes.HexBytes"];
+  // event nonce of the batch tx on Cosmos
+  uint64 nonce = 2;
+  // ethereum hex address of the contract
+  string token_contract = 3;
+}
+
+// LogicCallExecutedEvent describes a logic call that has been
+// successfully executed on Ethereum.
+message LogicCallExecutedEvent {
```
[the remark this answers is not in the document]
jkilpatr: The logic for that is in #327 but I'm not sure how that would be merged into this PR effectively.

module/proto/gravity/v1/gravity.proto
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// EthSignerSet is the Ethereum Bridge multisig set that relays transactions
+// the two chains. The staking validators keep ethereum keys which are used to
+// check signatures on Ethereum in order to get significant gas savings.
+message EthSignerSet {
+  // set of signers. Sorted by power
+  repeated EthSigner signers = 1 [(gogoproto.nullable) = false];
+  // TODO: which height? cosmos? This should be the key
+  uint64 height = 2;
```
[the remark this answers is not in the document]
jkilpatr: cosmos height.

module/x/gravity/keeper/batch.go
```diff
@@ -1,300 +1,362 @@
...
+	// cleanup outgoing transfer tx pool
+	for _, transferTxID := range batchTx.Transactions {
+		k.DeleteTransferTx(ctx, transferTxID)
 	}
-	var err error
-	// Iterate through remaining batches
-	k.IterateOutgoingTXBatches(ctx, func(key []byte, iter_batch *types.OutgoingTxBatch) bool {
+
+	// Iterate through remaining batches with the same token contract
+	// TODO: still not getting why we need to cancel the other batch txs?
```
jtremback, line 112: It is impossible for an earlier batch to be executed because its nonce is too low

module/x/gravity/keeper/batch.go
```diff
@@ -1,300 +1,362 @@
...
+		)
 	}
 
-	lastBatch := k.GetLastOutgoingBatchByTokenType(ctx, contractAddress)
+	timeoutHeight, err := k.GetBatchTimeoutHeight(ctx)
+	if err != nil {
+		return nil, err
+	}
 
-	// lastBatch may be nil if there are no existing batches, we only need
```
jtremback, line 33: Why remove the logic which ensures that only more profitable batches are created? Has it been moved somewhere else in the code? I've attached documentation of the feature that this PR appears to remove:

> Only create more profitable batches

> This is an important design criteria for permission-less batch creation.

> Suppose we have an infinite stream of low fee spam transactions going
into the Gravity tx pool. Any given batch creation event will then lock
up high fee 'good transactions' with dozens of spam transactions.

> In this case no relayer will ever chose to relay a batch because all of
them will be unprofitable.

> In order to ensure that profitable batches are eventually created we
must avoid locking up the high fee 'good transactions' into obviously
bad batches. To add to the difficulty we don't actually know what any
token in this process is worth or what ETH gas costs.

> The solution this patch provides is to ensure that a new batch always
has higher fees than the last batch. This means that even if there are
infinite spam transactions, and infinite spamming of the permission less
create batch request batch creation will halt long enough for enough
good transactions to build up in the pool and a new more profitable
batch to be created.

> Over a long enough timescale the old batches either time out. Returning
profitable transactions to the pool to create a new even better batch. Or
the profitability of the latest batch continues to increase until one is
relayed, also freeing the previous good transactions to go into the next
batch.

> So given this condition we can say that no matter the inputs a
successful batch will eventually be created.

jtremback: I searched for all the places this function is being called in the code, and it doesn't seem like the "only more profitable" logic was moved, it was simply removed. Why make key logic changes with no ADR? How will this affect the performance and spam resistance attributes of the bridge discussed above?

module/x/gravity/keeper/msg_server.go
```diff
@@ -7,137 +7,135 @@ import (
...
+		tokenContractHex := types.GravityDenomToERC20Contract(msg.Denom)
+		tokenContract = common.HexToAddress(tokenContractHex)
+	} else {
+		// get contract from store
+		tokenContract, found = k.GetERC20ContractFromCoinDenom(ctx, msg.Denom)
+		if !found {
+			// TODO: what if there is no corresponding contract? will it be "generated" on ethereum
+			// upon receiving?
+			// FIXME: this will fail if the cosmos tokens are relayed for the first time and they don't have a counterparty contract
+			// Fede: Also there's the question of how do we handle IBC denominations from a security perspective. Do we assign them the same
```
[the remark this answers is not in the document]
jtremback: Deploying an ERC20 on Ethereum costs a good amount of money (maybe a few hundred dollars?). This is why it is not done automatically. Someone with an interest in seeing a certain Cosmos token on Ethereum will have to pay the gas fee to make it happen. 

IBC denominations are all different tokens, and are not equivalent. They will be represented by different ERC20's on Ethereum. If a token comes over "SketchyChain", it is different from the same original token coming over the Hub. The SketchyChain relayed token will be valued less, reflecting the risk that the SketchyChain validators steal the underlying tokens, or whatever.

module/x/gravity/keeper/msg_server.go
```diff
@@ -2,175 +2,237 @@ package keeper
...
+			// contract? My guess is that each new contract assigned to a cosmos coin should be approved by governance
```
jtremback, line 155: This seems incredibly cumbersome

module/x/gravity/keeper/batch.go
jtremback: This logic is also discussed here: https://blog.althea.net/how-gravity-works

> As mentioned above, the Gravity module will only assemble a new batch if there are no batches already in the batch pool with a higher sum of fees. This is to prevent the following scenario: If there were many transactions in the transaction pool with an unprofitably low fee, and a few coming in every block with a high fee, each high fee transaction might end up in a batch with a bunch of unprofitable transactions. These batches would not be profitable to submit, and so the profitable transactions would end up in unprofitable batches, and not be submitted.

> By making it so that every new batch must be more profitable than any other batch that is waiting to be submitted, it gives the few profitable transactions that come in every block the chance to build up and form a batch profitable enough to submit.

module/x/gravity/keeper/batch.go
jtremback: This was also discussed here: https://blog.althea.net/how-gravity-works

>  After EthBlockDelay blocks on Ethereum, the validators submit the event to the Gravity module. When this event is fully observed (as detailed above), the batch is removed from the batch pool, since it is complete. Any batches of the same token with a lower batch nonce are also removed from the batch pool, but their transactions are also put back into the transaction pool. This is because we can be sure that it is impossible to submit them, since their batch nonces are too low. This mechanism allows transactions that were assembled into unprofitable batches to have a chance to make it into a profitable batch.

PR #321, opened 2021-04-11
Proto transactions for the Orchestrator

jkilpatr:
This pr includes proto transactions for the Orchestrator and must be merged after #308 as it is based off of that branch.

Proto transactions unblocks the refactor goal of merging all oracle messages into a single interface bound type and also ensures more reliable orchestrator operation since it's much easier to detect failures when using the proto endpoitns than the legacy amino endpoints.


zmanian (approved):
Most of the logic changes are in `deep_space`. But so happy to be finally unblocked here.

PR #322, opened 2021-04-11
WIP: Allow users to request arbitrary logic transactions

jkilpatr:
One of the current issues we have in repo is that the arbitrary logic testing flow is insufficient. There's no end to end test of arbitrary logic functionality and in fact some of the event parsing isn't tested at all.

The problem is that triggering arbitrary logic right now requires an external Cosmos module, so there's no good way to simulate an external module without actually having one.

I believe that we should expand the feature set of the Gravity bridge to allow arbitrary logic calls from regular users, which should be secure so long as we properly remove the required funds from the users accounts before allowing the call. We can then test this flow end to end in this repo and make sure that the vast majority of arbitrary logic functions work.

Right now this PR only removes the ill conceived testing end blocker, which has been disabled for some time anyways.

### DIFF
--- a/module/x/gravity/abci.go
+++ b/module/x/gravity/abci.go
@@ -266,31 +266,3 @@ func BatchSlashing(ctx sdk.Context, k keeper.Keeper, params types.Params) {
 
 	}
 }
-
-// TestingEndBlocker is a second endblocker function only imported in the Gravity codebase itself
-// if you are a consuming Cosmos chain DO NOT IMPORT THIS, it simulates a chain using the arbitrary
-// logic API to request logic calls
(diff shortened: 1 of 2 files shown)


zmanian:
I support removing this but I don't have a good idea for testing arbitrary logic in this repo.

jtremback (commented):
LGTM

jackzampolin:
I think we should not expose this functionality to users. Having a route for testing potentially OK, but also good might be a mock SDK module for testing.

One note on testing. I spend the weekending spiking on a go testing framework for spinning up and managing sdk clusters that would be extensible for use with gravity

PR #323, opened 2021-04-11
Compute normalized power difference

jkilpatr:
The Gravity contract depends on 'normalized' powers. The total amount of Cosmos voting power is summed and divided by 2^32 (uint32 max) and then normalized powers are used for voting.

We've had issues on testnets with the validator set generation code because it uses an absolute ratio of powers change. Which can in some cases **always** be greater than 1% or even 5%. For example if there is very high inflation it's possible that 5% of all staked tokens are actually issued every block. Resulting in the constant creation of new validator sets even though in terms of Gravity bridge normalized powers almost no power has been re-assigned.

This patch changes the power diff function to operate on normalized power differences, like the bridge itself does. This will hopefully cut down on the number of spurious validator set creations.

As a note, epoched staking will not remove the need for a power diff function. On the Cosmos hub for example normalized power changes at the rate of only a few percent a week. Assuming a reasonably sized epoch of 1000 blocks only relying on the Epoch will generate 121 validator sets versus just one which we can ballpark at an extra $5k to relay.

### DIFF
--- a/module/x/gravity/types/types.go
+++ b/module/x/gravity/types/types.go
@@ -64,14 +64,24 @@ func (b BridgeValidators) Sort() {
 }
 
 // PowerDiff returns the difference in power between two bridge validator sets
-// TODO: this needs to be potentially refactored
+// note this is Gravity bridge power *not* Cosmos voting power. Cosmos voting
+// power is based on the absolute number of tokens in the staking pool at any given
+// time Gravity bridge power is normalized using the equation.
+//
+// validators cosmos voting power / total cosmos voting power in this block = gravity bridge power / u32_max
+//
+// As an example if someone has 52% of the Cosmos voting power when a validator set is created their Gravity
+// bridge voting power is u32_max * .52
+//
+// Normalized voting power dramatically reduces how often we have to produce new validator set updates. For example
+// if the total on chain voting power increases by 1% due to inflation, we shouldn't have to generate a new validator
+// set, after all the validators retained their relative percentages during inflation and normalized Gravity bridge power
+// shows no difference.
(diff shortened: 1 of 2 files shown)


zmanian:
This change looks correct to me.

Could we get a comment in the code that capture analysis in the pull request?

fedekunze (dismissed):
ACK. Blocking on @zmanian's request for code comments

fedekunze:
why are we using power percentage instead of just the consensus power value?

jkilpatr:
> why are we using power percentage instead of just the consensus power value?

The consensus power value is constantly changing depending on inflation and the absolute value of tokens staked. The gravity bridge really doesn't need that info to accurately perform it's voting calculations. We could sum up all the powers and check for 66% of that, but it's just extra math to pay for forever on the Ethereum side for no actual advantage.

Note that the power voting threshold is just a deployment constant here

https://github.com/cosmos/gravity-bridge/blob/main/solidity/contracts/Gravity.sol#L42

we should probably just hardcode that to unint32max and save ourselves a storage lookup even.

jackzampolin (approved):

jkilpatr:
> ACK. Blocking on @zmanian's request for code comments

resolved code comments.

zmanian (approved):

jtremback (commented):
Informative

jackzampolin:
One note: in epoch an epoch world we would just check this once an epoch and only create a new one according to the power difference metric above.

jtremback (approved):

jkilpatr:
@fedekunze  could you update your review? I have resolved this and added comments.

PR #325, opened 2021-04-13
Documentation for relayer semantics and expansion of overview.md

jkilpatr:
This pull request contains some new documentation. There's a lot of docs to write so I'm starting with the things that are top of mind. If there's some specific documentation you would like let me know.

This pr is complete enough on it's own to merge, or could be expanded with more docs.

### DIFF
--- a/design/relaying-semantics.md
+++ b/design/relaying-semantics.md
@@ -1,0 +1,48 @@
+# Gravity bridge relaying semantics
+
+This document is designed to assist developers in implementing alternate Gravity relayers. The two major components of the Orchestrator which interact with Ethereum. The Gravity bridge has been designed for increased efficiency, not for ease of use. This means there are many implicit requirements of these external binaries which this document does it's best to make explicit.
+
+The Gravity `orchestrator` is described in [overview.md](overview.md) it's a combination of three distinct roles that need to be performed by external binaries in the Gravity bridge. This document highlights the requirements of the `relayer` which is one of those roles included in the `orchestrator`.
+
+## Semantics for Validator set update relaying
+
+### Sorting and Ordering of the Validator set and signatures
+
+When updating the validator set in the Gravity contract you must provide a copy of the old validator set. This _MUST_ only be taken from the last ValsetUpdated event on the Ethereum chain.
+
+Providing the old validator set is part of a storage optimization, instead of storing the entire validator set in Ethereum storage it is instead provided by each caller and stored in the much cheaper Ethereum event queue. No sorting of any kind is performed in the Gravity contract, meaning the list of validators and their new signatures must be submitted in exactly the same order as the last call.
+
+For the purpose of normal operation this requirement can be shortened to 'sort the validators by descending power, and by Eth address bytes where power is equal'. Since the Cosmos module produces the validator sets they should always come in order. But a flaw in this sorting method that caused an unsorted validator set to make it's way into the chain would halt valset updates and essentially decouple the bridge unless your implementation is smart enough to take a look at the last submitted order rather than blindly following sorting.
+
+### Deciding what Validator set to relay
+
+The Cosmos chain simply produces a stream of validator sets, it does not make any judgement on how they are relayed. It's up to the relayer implementation to determine how to optimize the gas costs of this relaying operation.
+
+For example lets say we had validator sets `A, B, C, and D` each is created when there is a 5% power difference between the last Gravity validator set snapshot in the store and the currently active validator set.
+
+5% is an arbitrary constant. The specific value chosen here is a tradeoff made by the chain between how up to date the Ethereum validator set is and the cost to keep it updated. The higher this value is the lower the portion of the voting validator set is needed to highjack the bridge in the worst case. If we made a new validator set update every block 66% would need to collude, the 5% change threshold means 61% of the total voting power colluding in a given validator set may be able to steal the funds in the bridge.
+
+```
+A -> B -> C -> D
+     5%  10%   15%
+```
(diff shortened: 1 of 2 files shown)


design/relaying-semantics.md
```diff
@@ -0,0 +1,44 @@
+# Gravity bridge relaying semantics
+
+This document is designed to assist developers in implementing alternate Gravity relayers. The two major components of the Orchestrator which interact with Ethereum. The Gravity bridge has been designed for increased efficiency, not for ease of use. This means there are many implicit requirements of these external binaries which this document does it's best to make explicit.
```
marbar3778, line 3: can you mention what the orchestrator is, it may be beneficial to also avoid using this term within this section as the relayer is only part of the orchestrator.

design/relaying-semantics.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Semantics for Validator set update relaying
+
+### Sorting and Ordering of the Validator set and signatures
+
+When updating the validator set in the Gravity contact you must provide a copy of the old validator set. This _MUST_ only be taken from the last ValsetUpdated event on the Ethereum chain.
```
marbar3778, line 9:
```suggestion
When updating the validator set in the Gravity contract you must provide a copy of the old validator set. This _MUST_ only be taken from the last ValsetUpdated event on the Ethereum chain.
```

design/relaying-semantics.md
```diff
@@ -0,0 +1,44 @@
...
+
+For the purpose of normal operation this requirement can be shortened to 'sort the validators by descending power, and by Eth address bytes where power is equal'. Since the Cosmos module produces the validator sets they should always come in order. But a flaw in this sorting method that caused an unsorted validator set to make it's way into the chain would halt valset updates and essentially decouple the bridge unless your implementation is smart enough to take a look at the last submitted order rather than blindly following sorting.
+
+### Deciding what Validator set to relay
+
+The Cosmos chain simply produces a stream of validator sets, it does not make any judgement on how they are relayed. It's up to the relayer implementation to determine how to optimize the gas costs of this relaying operation.
+
+For example lets say we had validator sets `A, B, C, and D` each is created when there is a 5% power difference between the last Gravity validator set snapshot in the store and the currently active validator set.
```
marbar3778, line 19: how was 5% chosen? maybe add some words to this

marbar3778 (approved):
left some minor notes, but looks good

design/relaying-semantics.md
jkilpatr: that's a good call. I'll spend some more time disambiguating it.

design/relaying-semantics.md
jkilpatr: well it was 1% but that generated too many. But yes it's arbitrary and should probably be brought back down to 1% since we've root caused the issue in #323

design/relaying-semantics.md
```diff
@@ -0,0 +1,48 @@
...
+
+When updating the validator set in the Gravity contract you must provide a copy of the old validator set. This _MUST_ only be taken from the last ValsetUpdated event on the Ethereum chain.
+
+Providing the old validator set is part of a storage optimization, instead of storing the entire validator set in Ethereum storage it is instead provided by each caller and stored in the much cheaper Ethereum event queue. No sorting of any kind is performed in the Gravity contract, meaning the list of validators and their new signatures must be submitted in exactly the same order as the last call.
+
+For the purpose of normal operation this requirement can be shortened to 'sort the validators by descending power, and by Eth address bytes where power is equal'. Since the Cosmos module produces the validator sets they should always come in order. But a flaw in this sorting method that caused an unsorted validator set to make it's way into the chain would halt valset updates and essentially decouple the bridge unless your implementation is smart enough to take a look at the last submitted order rather than blindly following sorting.
```
jackzampolin, line 15: Is this something that could be done by an adversarial relayer? I.e. take the signed message from the chain and reorder the sigs?

design/relaying-semantics.md
```diff
@@ -0,0 +1,48 @@
...
+
+Note that `eth_call()` often has funny gotchas. All calls fail on Geth based implementations if you don't have any Ethereum to pay for gas, while on Parity based implementations your gas inputs are mostly ignored and an accurate gas usage is returned.
+
+## Semantics for transaction batch relaying
+
+In order to submit a transaction batch you also need to submit the last set of validators and their powers as outlined in [the validator set section](### Sorting and Ordering of the Validator set and signatures). This is to facilitate the same storage optimization mentioned there.
+
+### Deciding what batch to relay
+
+Making a decision about which batch to relay is very different from deciding which validator set to relay. Batch relaying is primarily motivated by fees, not by a desire to maintain the integrity of the bridge. So the decision mostly comes down to fee computation, this is further complicated by the concept of 'batch requests'. Which is an unpermissioned transaction that requests the Gravity module generate a new batch for a specific token type.
```
jackzampolin, line 42: I propose that we permission this transaction to be only callable by the validators or orchestrators

jkilpatr: My reasoning for why it's safe to allow anyone to call this is covered here

https://github.com/cosmos/gravity-bridge/blob/main/spec/batch-creation-spec.md

I'm interested in your reasoning for permissioning the call given the logic described in the above doc?

design/relaying-semantics.md
jkilpatr: As we discussed in Telegram this isn't really doable and I should open up a new pull to clarify. Thanks for pointing out that this is confusing.

PR #326, opened 2021-04-13
Validator set pruning using relayed Validator sets

jkilpatr:
### Reasoning

This pull request re-introduces pruning for validator sets but in a way that is guaranteed to be secure.

In the past we simply pruned validator sets after a given number of blocks. This solution was simple and would work well the vast majority of the time. But could result in a catastrophic failure of the bridge in some edge cases.

Lets take for example a case where the EVM chain halts for 24 hours for whatever reason. During this time the Cosmos chain continues normally and has a complete validator set change out with no hostile intent on the part of any validator. The former pruning logic would remove the signed validator sets from the history as the slashing window passed and by the time the EVM chain restarted no update signed by the validator set that controlled the chain 24 hours ago would still be in the store.

The bridge and all the funds in it would be lost at this point. Even though the signatures required to transition it to the new validator set where created and submitted, they where just deleted before they could be used.

The only solution to this edge case is to do something we've wanted to avoid for a while. Ferry the ValsetUpdated event from Ethereum to Cosmos as an Oracle claim.

We have tried to avoid this mainly because when it was around early in the development of Gravity developers had a habit of misusing it and generally assuming it's more accurate than is possible considering the constraints imposed by the separate timing of two different chains. 

But without some awareness of the past state of the Ethereum chain we would have to wait weeks before we could safely assume that the validator set had been updated on the EVM chain. This was previously accepted as a 'solution' but the failure of testnet1 under mountains of validator set updates is a pretty strong takedown of that design decision.

### Secondary Reasoning

In addition to our pruning problem having a reasonable idea of what validator set exists on Ethereum will resolve some audit issues. One of the audit issues Informal raised was that a Cosmos full node operator might not know that the bridge has been highjacked.

Now that we are ferrying the data we can (although this pull request does not implement it) validate that the validator set on Ethereum matches the stored value for that nonce on Cosmos. In the case of a failure the Cosmos chain can panic, or raise whatever warning is required without requiring the user to run any sort of external binary to determine that.


jkilpatr:
jkilplatr nice, closing to fix my branch name.

PR #327, opened 2021-04-13
Validator set pruning using relayed Validator sets

jkilpatr:
### Reasoning

This pull request re-introduces pruning for validator sets but in a way that is guaranteed to be secure.

In the past we simply pruned validator sets after a given number of blocks. This solution was simple and would work well the vast majority of the time. But could result in a catastrophic failure of the bridge in some edge cases.

Lets take for example a case where the EVM chain halts for 24 hours for whatever reason. During this time the Cosmos chain continues normally and has a complete validator set change out with no hostile intent on the part of any validator. The former pruning logic would remove the signed validator sets from the history as the slashing window passed and by the time the EVM chain restarted no update signed by the validator set that controlled the chain 24 hours ago would still be in the store.

The bridge and all the funds in it would be lost at this point. Even though the signatures required to transition it to the new validator set where created and submitted, they where just deleted before they could be used.

The only solution to this edge case is to do something we've wanted to avoid for a while. Ferry the ValsetUpdated event from Ethereum to Cosmos as an Oracle claim.

We have tried to avoid this mainly because when it was around early in the development of Gravity developers had a habit of misusing it and generally assuming it's more accurate than is possible considering the constraints imposed by the separate timing of two different chains. 

But without some awareness of the past state of the Ethereum chain we would have to wait weeks before we could safely assume that the validator set had been updated on the EVM chain. This was previously accepted as a 'solution' but the failure of testnet1 under mountains of validator set updates is a pretty strong takedown of that design decision.

### Secondary Reasoning

In addition to our pruning problem having a reasonable idea of what validator set exists on Ethereum will resolve some audit issues. One of the audit issues Informal raised was that a Cosmos full node operator might not know that the bridge has been highjacked.

Now that we are ferrying the data we can (although this pull request does not implement it) validate that the validator set on Ethereum matches the stored value for that nonce on Cosmos. In the case of a failure the Cosmos chain can panic, or raise whatever warning is required without requiring the user to run any sort of external binary to determine that.


jackzampolin:
Like this change. cc @fedekunze @marbar3778 this will need to be in the refactor.

jackzampolin (approved):
This is a great change!

jackzampolin:
Looks like integration tests are still broken.

marbar3778 (changes requested):
This needs documentation within the PR.

module/x/gravity/abci.go
```diff
@@ -10,16 +10,20 @@ import (
...
 func EndBlocker(ctx sdk.Context, k keeper.Keeper) {
+	params := k.GetParams(ctx)
 	// Question: what here can be epoched?
 	slashing(ctx, k)
 	attestationTally(ctx, k)
 	cleanupTimedOutBatches(ctx, k)
 	cleanupTimedOutLogicCalls(ctx, k)
 	createValsets(ctx, k)
+	pruneValsets(ctx, k, params)
+	// TODO: prune claims, attestations when they pass in the handler
```
marbar3778, line 21: please open a issue.

module/x/gravity/keeper/attestation_handler.go
```diff
@@ -111,6 +111,14 @@ func (a AttestationHandler) Handle(ctx sdk.Context, att types.Attestation, claim
 
 		// Add to denom-erc20 mapping
 		a.keeper.setCosmosOriginatedDenomToERC20(ctx, claim.CosmosDenom, claim.TokenContract)
+	case *types.MsgValsetUpdatedClaim:
+		// TODO here we should check the contents of the validator set against
```
marbar3778, line 115: please open an issue.

module/x/gravity/keeper/msg_server.go
```diff
@@ -439,6 +439,46 @@ func (k msgServer) LogicCallExecutedClaim(c context.Context, msg *types.MsgLogic
...
+	_, err = k.Attest(ctx, msg, any)
+	if err != nil {
+		return nil, sdkerrors.Wrap(err, "create attestation")
+	}
+
+	// Emit the handle message event
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, msg.Type()),
```
marbar3778, line 473: The message type is not the module.

marbar3778 (commented):
overall looks good. It would be good to capture the design in a markdown file, this will help future teams with onboarding.

module/x/gravity/abci.go
jkilpatr: #342

module/x/gravity/keeper/attestation_handler.go
jkilpatr: #343

jkilpatr:
> overall looks good. It would be good to capture the design in a markdown file, this will help future teams with onboarding.

@jtremback  is working on the spec here.

module/x/gravity/keeper/msg_server.go
jkilpatr: this is generally the pattern we use in the file. So if this is an issue we should open an issue and change it throughout the fie.

PR #328, opened 2021-04-15
Fix MsgCancelSendToEth

jkilpatr:
It seems I was in a bit too much of a rush when copying the signing
boilerplate for this message type.

Note that I am relying on the get_signers implementation to ensure
msg.Sender has signed the message. In msg_server.go we just remove the
tx from the pool with the assumption that it's from the sender.

thanks to @mankenavenkatesh  for pointing this out.

### DIFF
--- a/module/x/gravity/keeper/pool.go
+++ b/module/x/gravity/keeper/pool.go
@@ -101,6 +101,16 @@ func (k Keeper) RemoveFromOutgoingPoolAndRefund(ctx sdk.Context, txId uint64, se
 		return err
 	}
 
+	// Check that this user actually sent the transaction, this prevents someone from refunding someone
+	// elses transaction to themselves.
+	txSender, err := sdk.AccAddressFromBech32(tx.Sender)
+	if err != nil {
+		panic("Invalid address in store!")
+	}
+	if !txSender.Equals(sender) {
+		return sdkerrors.Wrapf(types.ErrInvalid, "Sender %s did not send Id %d", sender, txId)
+	}
+
 	found := false
 	poolTx := k.GetPoolTransactions(ctx)
 	for _, pTx := range poolTx {
--- a/module/x/gravity/types/msgs.go
+++ b/module/x/gravity/types/msgs.go
@@ -13,6 +13,7 @@ var (
 	_ sdk.Msg = &MsgSetOrchestratorAddress{}
 	_ sdk.Msg = &MsgValsetConfirm{}
 	_ sdk.Msg = &MsgSendToEth{}
+	_ sdk.Msg = &MsgCancelSendToEth{}
 	_ sdk.Msg = &MsgRequestBatch{}
 	_ sdk.Msg = &MsgConfirmBatch{}
 	_ sdk.Msg = &MsgERC20DeployedClaim{}
@@ -616,8 +617,9 @@ func (b *MsgValsetUpdatedClaim) ClaimHash() []byte {
 }
 
 // NewMsgCancelSendToEth returns a new msgSetOrchestratorAddress
-func NewMsgCancelSendToEth(val sdk.ValAddress, id uint64) *MsgCancelSendToEth {
+func NewMsgCancelSendToEth(user sdk.AccAddress, id uint64) *MsgCancelSendToEth {
 	return &MsgCancelSendToEth{
+		Sender:        user.String(),
 		TransactionId: id,
 	}
 }
@@ -630,7 +632,7 @@ func (msg *MsgCancelSendToEth) Type() string { return "cancel_send_to_eth" }
 
 // ValidateBasic performs stateless checks
 func (msg *MsgCancelSendToEth) ValidateBasic() (err error) {
-	_, err = sdk.ValAddressFromBech32(msg.Sender)
+	_, err = sdk.AccAddressFromBech32(msg.Sender)
 	if err != nil {
 		return err
 	}
@@ -644,11 +646,11 @@ func (msg *MsgCancelSendToEth) GetSignBytes() []byte {
 
 // GetSigners defines whose signature is required
 func (msg *MsgCancelSendToEth) GetSigners() []sdk.AccAddress {
-	acc, err := sdk.ValAddressFromBech32(msg.Sender)
+	acc, err := sdk.AccAddressFromBech32(msg.Sender)
 	if err != nil {
 		panic(err)
 	}
-	return []sdk.AccAddress{sdk.AccAddress(acc)}
+	return []sdk.AccAddress{acc}
 }
 
 // MsgSubmitBadSignatureEvidence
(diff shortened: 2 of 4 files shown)


mankenavenkatesh:
@jkilpatr 
We also need to check if the sender on the MsgSendToEth is the signer of the cancel transaction. 
Currently anyone can send MsgCancelSendToEth with my tx id.

jkilpatr:
> @jkilpatr
> We also need to check if the sender on the MsgSendToEth is the signer of the cancel transaction.
> Currently anyone can send MsgCancelSendToEth with my tx id.

Please check the design

https://github.com/cosmos/gravity-bridge/blob/main/module/x/gravity/keeper/msg_server.go#L448

here we take the sender from the message

Here we make sure the address specified by sender signs the message

https://github.com/cosmos/gravity-bridge/blob/main/module/x/gravity/types/msgs.go#L579

Therefore no one can send MsgCancelSendToEth for someone else, only for themselves, otherwise they don't meet the signer requirements.

mankenavenkatesh:
@jkilpatr Let's go through an example

**Problem**
1. First,  UserA sent `MsgSendToEth` tx. we have deducted UserA balance and created OutgoingTransferTx.**(txID = 1)**

2. Then, UserB signs and send `MsgCancelSendToEth` tx with following contents where sender is UserB but **txID = 1**.
```
 MsgCancelSendToEth {
	TransactionId : 1,
	Sender : userB,
}
```

3. During signature validation of `MsgCancelSendToEth`,  as signer and sender are both userB, the validation is successful. 
https://github.com/cosmos/gravity-bridge/blob/a9e2e31efc12dbd068886cba35316245d5dc229a/module/x/gravity/types/msgs.go#L579

4. Next when executing `MsgCancelSendToEth`, we fetch **txID = 1** details and transfer tokens to sender (In this example to userB). 
https://github.com/cosmos/gravity-bridge/blob/a9e2e31efc12dbd068886cba35316245d5dc229a/module/x/gravity/keeper/pool.go#L99

5. In summary, userB is able cancel and takeover tokens of the withdrawTx of userA.


--------

**Solution**
Add following check in `RemoveFromOutgoingPoolAndRefund` function
```
	txSender, err := sdk.AccAddressFromBech32(tx.Sender)
	if err != nil {
		return err
	}

	if !sender.Equals(txSender) {
		return sdkerrors.Wrapf(types.ErrInvalid, "Invalid sender address")
	}
```

jkilpatr:
> @jkilpatr Let's go through an example
> 
> **Problem**
> 
>     1. First,  UserA sent `MsgSendToEth` tx. we have deducted UserA balance and created OutgoingTransferTx.**(txID = 1)**
> 
>     2. Then, UserB signs and send `MsgCancelSendToEth` tx with following contents where sender is UserB but **txID = 1**.
> 
> 
> ```
>  MsgCancelSendToEth {
> 	TransactionId : 1,
> 	Sender : userB,
> }
> ```
> 
>     1. During signature validation of `MsgCancelSendToEth`,  as signer and sender are both userB, the validation is successful.
>        https://github.com/cosmos/gravity-bridge/blob/a9e2e31efc12dbd068886cba35316245d5dc229a/module/x/gravity/types/msgs.go#L579
> 
>     2. Next when executing `MsgCancelSendToEth`, we fetch **txID = 1** details and transfer tokens to sender (In this example to userB).
>        https://github.com/cosmos/gravity-bridge/blob/a9e2e31efc12dbd068886cba35316245d5dc229a/module/x/gravity/keeper/pool.go#L99
> 
>     3. In summary, userB is able cancel and takeover tokens of the withdrawTx of userA.
> 
> 
> **Solution**
> Add following check in `RemoveFromOutgoingPoolAndRefund` function
> 
> ```
> 	txSender, err := sdk.AccAddressFromBech32(tx.Sender)
> 	if err != nil {
> 		return err
> 	}
> 
> 	if !sender.Equals(txSender) {
> 		return sdkerrors.Wrapf(types.ErrInvalid, "Invalid sender address")
> 	}
> ```

You're correct. I forgot that the txid's are global. I'll upload a better fix some time today.

jkilpatr:
@mankenavenkatesh  I've added your suggestion thanks 👍🏻

jackzampolin (approved):

module/x/gravity/keeper/pool.go
```diff
@@ -101,6 +101,16 @@ func (k Keeper) RemoveFromOutgoingPoolAndRefund(ctx sdk.Context, txId uint64, se
 		return err
 	}
 
+	// Check that this user actually sent the transaction, this prevents someone from refunding someone
+	// elses transaction to themselves.
+	txSender, err := sdk.AccAddressFromBech32(tx.Sender)
```
marbar3778, line 106: could we add a regression test here.

tac0turtle (dismissed):
Could we get a regression test or a test that follows the behavior mentioned in the above comment?

module/x/gravity/types/msgs.go
```diff
@@ -549,8 +549,9 @@ func (b *MsgLogicCallExecutedClaim) ClaimHash() []byte {
 }
 
 // NewMsgSetOrchestratorAddress returns a new msgSetOrchestratorAddress
-func NewMsgCancelSendToEth(val sdk.ValAddress, id uint64) *MsgCancelSendToEth {
+func NewMsgCancelSendToEth(user sdk.AccAddress, id uint64) *MsgCancelSendToEth {
```
marbar3778, line 552: Slightly confused, this message isnt registered on the handler, how is this suppose to be executed?

fedekunze: lol it won't. I noticed it and fixed on my PR

module/x/gravity/types/msgs.go
```diff
@@ -577,9 +578,9 @@ func (msg *MsgCancelSendToEth) GetSignBytes() []byte {
 
 // GetSigners defines whose signature is required
 func (msg *MsgCancelSendToEth) GetSigners() []sdk.AccAddress {
-	acc, err := sdk.ValAddressFromBech32(msg.Sender)
+	acc, err := sdk.AccAddressFromBech32(msg.Sender)
```
marbar3778, line 581: This changes the original behavior, the bech32 prefixes are different for these two addresses. We should double-check how txs are submitted here to make sure it's correct.

jackzampolin: ideally we standardize this behavior across the app. I'm in favor of only using the accaddress prefix for validators

fedekunze: i fixed this on my refactor

module/x/gravity/types/msgs.go
jkilpatr: corrected this here.

jkilpatr:
> Could we get a regression test or a test that follows the behavior mentioned in the above comment?

what file should that go in? since the message handler got moved from types/msgs.go to keeper/msg_server.go I haven't been sure if I should add a message server test or use msgs_test in types.

marbar3778:
> what file should that go in? since the message handler got moved from types/msgs.go to keeper/msg_server.go I haven't been sure if I should add a message server test or use msgs_test in types.

message_server seems like a better place. Amino will be deprecated shortly.

module/x/gravity/keeper/pool.go
jtremback: @marbar3778 Added

PR #332, opened 2021-04-17
Add support for custom prefixes

jkilpatr:
This patch adds support for custom prefixes. To allow for chains with
different address prefixes than the default 'cosmos'.

Sadly this requires the user to specify the prefix as an argument when
launching the orchestrator. There's no way around this until a future
version of Cosmos exposes the prefix to clients. Looking upstream the
work for this seems to be in progress.

### DIFF
--- a/orchestrator/gravity_utils/src/connection_prep.rs
+++ b/orchestrator/gravity_utils/src/connection_prep.rs
@@ -26,6 +26,7 @@ pub struct Connections {
 /// operation in a error resilient manner. TODO find some way to generalize
 /// this so that it's less ugly
 pub async fn create_rpc_connections(
+    chain_prefix: String,
     grpc_url: Option<String>,
     eth_rpc_url: Option<String>,
     timeout: Duration,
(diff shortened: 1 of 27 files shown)


orchestrator/relayer/src/main.rs
```diff
@@ -25,17 +25,19 @@ extern crate log;
...
     flag_cosmos_grpc: String,
+    flag_cosmos_prefix: String,
     flag_ethereum_rpc: String,
     flag_contract_address: String,
 }
 
 lazy_static! {
     pub static ref USAGE: String = format!(
-    "Usage: {} --ethereum-key=<key> --cosmos-grpc=<url> --ethereum-rpc=<url> --contract-address=<addr>
+    "Usage: {} --ethereum-key=<key> --cosmos-grpc=<url> --cosmos-prefix=<prefix> --ethereum-rpc=<url> --contract-address=<addr>
```
marbar3778, line 35:
```suggestion
    "Usage: {} --ethereum-key=<key> --cosmos-grpc=<url> --address-prefix=<prefix> --ethereum-rpc=<url> --contract-address=<addr>
```

Should this be `address-prefix`, cosmos-prefix is broad.

orchestrator/gravity_utils/src/connection_prep.rs
```diff
@@ -26,6 +26,7 @@ pub struct Connections {
 /// operation in a error resilient manner. TODO find some way to generalize
 /// this so that it's less ugly
 pub async fn create_rpc_connections(
+    chain_prefix: String,
```
marbar3778, line 29: Can we name this address_prefix and do so throughout the PR. cosmos_prefix and chain_prfix are a bit confusing

jtremback (approved):

orchestrator/gravity_utils/src/connection_prep.rs
fedekunze: the correct term would be `bech32_hrp` (human readable prefix)

fedekunze (approved):
utACK. Is the default prefix set to `cosmos`? ideally, the flag should be an optional

jkilpatr:
> utACK. Is the default prefix set to `cosmos`? ideally, the flag should be an optional

default is Cosmos, but that's in the library this code imports. Not here.

orchestrator/gravity_utils/src/connection_prep.rs
jkilpatr: rename done. We'll be able to remove it as an argument pretty soon. There's upstream work to allow prefixes to be simply queried by the library.

PR #336, opened 2021-04-28
Evidence slashing

jtremback:
This PR adds an endpoint which allows someone to submit a batch, valset, or logic call, (the "subject") and the signature of a validator. 

It checks the subject's checkpoint against a stored set of all past checkpoints, and if it cannot find the checkpoint, the validator who made the signature is slashed. This is intended to deter validators from signing fake batches, valsets, or logic calls, for example if they wanted to steal money from the bridge with a fake transaction.


module/proto/gravity/v1/genesis.proto
```diff
@@ -99,6 +99,10 @@ message Params {
     (gogoproto.nullable)   = false
   ];
   uint64 unbond_slashing_valsets_window = 17;
+  bytes slash_fraction_bad_eth_signature = 18 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
+    (gogoproto.nullable)   = false
+  ];
```
fedekunze, line 105: this needs to be a string. Also, what does a "bad" signature mean in this context

module/proto/gravity/v1/msgs.proto
```diff
@@ -235,3 +239,13 @@ message MsgCancelSendToEth {
...
 
 message MsgCancelSendToEthResponse {}
+
+// This call allows anyone to submit evidence that a
+// validator has signed a valset, batch, or logic call that never
+// existed. Subject contains the batch, valset, or logic call.
+message MsgSubmitBadSignatureEvidence {
+  google.protobuf.Any subject   = 1;
+  string              signature = 2;
+}
```
fedekunze, line 249: rename to `MsgSubmitMisbehavior`, `subject` is also a bad term here and should be included in the misbehavior field. This msg should also contain the signer account address

module/x/gravity/abci.go
```diff
@@ -156,7 +157,7 @@ func ValsetSlashing(ctx sdk.Context, k keeper.Keeper, params types.Params) {
 				// Check if validator has confirmed valset or not
 				found := false
 				for _, conf := range confirms {
-					if conf.EthAddress == k.GetEthAddress(ctx, val.GetOperator()) {
+					if conf.EthAddress == k.GetEthAddressByValidator(ctx, val.GetOperator()) {
```
fedekunze, line 160: change out of scope

module/x/gravity/keeper/batch.go
```diff
@@ -59,6 +59,10 @@ func (k Keeper) BuildOutgoingTXBatch(
 	}
 	k.StoreBatch(ctx, batch)
 
+	// Get the checkpoint and store it as a legit past batch
+	checkpoint := batch.GetCheckpoint(k.GetGravityID(ctx))
+	k.SetPastEthSignatureCheckpoint(ctx, checkpoint)
```
fedekunze, line 64: how is this related to the evidence?

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+	"encoding/hex"
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/cosmos/gravity-bridge/module/x/gravity/types"
+)
+
+func (k Keeper) CheckBadSignatureEvidence(
```
fedekunze, line 14: explain what this function does. Again, bad signature is not clear.

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+	ctx sdk.Context,
+	msg *types.MsgSubmitBadSignatureEvidence) error {
```
fedekunze, line 16: should be misbehavior instead of the full msg

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+
+	k.cdc.UnpackAny(msg.Subject, &subject)
+
+	switch subject := subject.(type) {
+	case *types.OutgoingTxBatch:
+		return k.checkBadSignatureEvidenceInternal(ctx, subject, msg.Signature)
+	case *types.Valset:
+		return k.checkBadSignatureEvidenceInternal(ctx, subject, msg.Signature)
+	case *types.OutgoingLogicCall:
+		return k.checkBadSignatureEvidenceInternal(ctx, subject, msg.Signature)
```
fedekunze, line 27: I don't understand why you do you have the switch in the first place if you are calling the same function?

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+
+	"github.com/cosmos/gravity-bridge/module/x/gravity/types"
+)
+
+func (k Keeper) CheckBadSignatureEvidence(
+	ctx sdk.Context,
+	msg *types.MsgSubmitBadSignatureEvidence) error {
+	var subject types.EthereumSigned
```
fedekunze, line 17: rename

module/x/gravity/keeper/batch.go
fedekunze: ~~also, why is this overriding the previous checkpoint? shouldn't you keep every checkpoint in-store instead of a single one?~~

See comment below about using the checkpoint as the key

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	// Slash the offending validator
+	cons, err := val.GetConsAddr()
+	if err != nil {
+		return sdkerrors.Wrap(err, "Could not get consensus key address for validator")
+	}
+
+	params := k.GetParams(ctx)
+	k.StakingKeeper.Slash(ctx, cons, ctx.BlockHeight(), val.ConsensusPower(), params.SlashFractionBadEthSignature)
```
fedekunze, line 70: FYI this panics if the validator is not bonded

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+	// Decode Eth signature to bytes
+	sigBytes, err := hex.DecodeString(signature)
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrInvalid, "signature decoding")
+	}
+
+	// Get eth address of the offending validator using the checkpoint and the signature
+	ethAddress, err := types.EthAddressFromSignature(checkpoint, sigBytes)
+	if err != nil {
+		return sdkerrors.Wrap(types.ErrInvalid, fmt.Sprintf("signature to eth address failed with checkpoint %s and signature %s", hex.EncodeToString(checkpoint), signature))
```
fedekunze, line 54: use wrapf

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (k Keeper) checkBadSignatureEvidenceInternal(ctx sdk.Context, subject types.EthereumSigned, signature string) error {
+	// Get checkpoint of the supposed bad signature (fake valset, batch, or logic call submitted to eth)
+	gravityID := k.GetGravityID(ctx)
+	checkpoint := subject.GetCheckpoint(gravityID)
+
+	// Try to find the checkpoint in the archives. If it exists, we don't slash because
+	// this is not a bad signature
+	if k.GetPastEthSignatureCheckpoint(ctx, checkpoint) {
```
fedekunze, line 41: ditto, this will only fetch the latest stored checkpoint

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// GetPastEthSignatureCheckpoint tells you whether a given checkpoint has ever existed
+func (k Keeper) GetPastEthSignatureCheckpoint(ctx sdk.Context, checkpoint []byte) (found bool) {
+	store := ctx.KVStore(k.storeKey)
+	if bytes.Equal(store.Get(types.GetPastEthSignatureCheckpointKey(checkpoint)), []byte{0x1}) {
+		return true
+	} else {
+		return false
+	}
```
fedekunze, line 92:
```suggestion
	return bytes.Equal(store.Get(types.GetPastEthSignatureCheckpointKey(checkpoint)), []byte{0x1})
```

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,93 @@
...
+
+	return nil
+}
+
+// SetPastEthSignatureCheckpoint puts the checkpoint of a valset, batch, or logic call into a set
+// in order to prove later that it existed at one point.
+func (k Keeper) SetPastEthSignatureCheckpoint(ctx sdk.Context, checkpoint []byte) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.GetPastEthSignatureCheckpointKey(checkpoint), []byte{0x1})
```
fedekunze, line 82: why are you storing a byte as the value instead of the checkpoint?

module/x/gravity/keeper/evidence_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/gravity-bridge/module/x/gravity/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSubmitBadSignatureEvidenceBatchExists(t *testing.T) {
```
fedekunze, line 15: these are bad testing patterns... are you covering all cases? I'd suggest refactoring these as part of this PR

module/x/gravity/keeper/keeper.go
```diff
@@ -369,15 +373,31 @@ func (k Keeper) GetOrchestratorValidator(ctx sdk.Context, orch sdk.AccAddress) s
...
+	valAddr := store.Get(types.GetValidatorByEthAddressKey(ethAddr))
+	if valAddr == nil {
+		return stakingtypes.Validator{}, false
+	}
+	validator, found = k.StakingKeeper.GetValidator(ctx, valAddr)
+	if !found {
+		return stakingtypes.Validator{}, false
+	}
+
+	return validator, true
```
fedekunze, line 400: I'd highly recommend checking the Uber golang style guide for reference: https://github.com/uber-go/guide

```suggestion
	return k.StakingKeeper.GetValidator(ctx, valAddr)
```

module/x/gravity/keeper/keeper.go
```diff
@@ -641,7 +666,7 @@ func (k Keeper) UnpackAttestationClaim(att *types.Attestation) (types.EthereumCl
 // For the time being this will serve
 func (k Keeper) GetDelegateKeys(ctx sdk.Context) []*types.MsgSetOrchestratorAddress {
 	store := ctx.KVStore(k.storeKey)
-	prefix := []byte(types.EthAddressKey)
+	prefix := []byte(types.EthAddressByValidatorKey)
```
fedekunze, line 669: why this change?

module/x/gravity/keeper/msg_server.go
```diff
@@ -460,3 +454,20 @@ func (k msgServer) CancelSendToEth(c context.Context, msg *types.MsgCancelSendTo
...
+
+func (k msgServer) SubmitBadSignatureEvidence(c context.Context, msg *types.MsgSubmitBadSignatureEvidence) (*types.MsgSubmitBadSignatureEvidenceResponse, error) {
+	ctx := sdk.UnwrapSDKContext(c)
+
+	err := k.CheckBadSignatureEvidence(ctx, msg)
+
+	ctx.EventManager().EmitEvent(
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, msg.Type()),
```
fedekunze, line 466: this is not what the module attribute is for, it should be the module name

module/x/gravity/keeper/msg_server.go
```diff
@@ -460,3 +454,20 @@ func (k msgServer) CancelSendToEth(c context.Context, msg *types.MsgCancelSendTo
...
+			sdk.NewAttribute(types.AttributeKeyBadEthSignature, fmt.Sprint(msg.Signature)),
```
fedekunze, line 467: no need for fmt.Sprint

module/x/gravity/keeper/msg_server.go
```diff
@@ -460,3 +454,20 @@ func (k msgServer) CancelSendToEth(c context.Context, msg *types.MsgCancelSendTo
...
+			sdk.NewAttribute(types.AttributeKeyBadEthSignatureSubject, fmt.Sprint(msg.Subject)),
```
fedekunze, line 468: why do you need to emit an event with the entire data? also, not sure if you want to print this in a byte format like is currently doing

module/x/gravity/keeper/test_common.go
```diff
@@ -408,6 +410,9 @@ func CreateTestEnv(t *testing.T) TestInput {
 	)
 
 	k.SetParams(ctx, TestingGravityParams)
+	params := k.GetParams(ctx)
+
+	fmt.Println(params)
```
fedekunze, line 415: remove

module/x/gravity/types/batch.go
```diff
@@ -1,21 +1,21 @@
...
 
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
 // GetCheckpoint gets the checkpoint signature from the given outgoing tx batch
-func (b OutgoingTxBatch) GetCheckpoint(gravityIDstring string) ([]byte, error) {
+func (b OutgoingTxBatch) GetCheckpoint(gravityIDstring string) []byte {
```
fedekunze, line 14: are we expecting these checkpoints to panic now?

module/x/gravity/types/events.go
```diff
@@ -10,19 +10,21 @@ const (
...
+	AttributeKeyContract               = "bridge_contract"
+	AttributeKeyNonce                  = "nonce"
+	AttributeKeyValsetNonce            = "valset_nonce"
+	AttributeKeyBatchNonce             = "batch_nonce"
+	AttributeKeyBridgeChainID          = "bridge_chain_id"
+	AttributeKeySetOperatorAddr        = "set_operator_address"
+	AttributeKeyInvalidationID         = "logic_call_invalidation_id"
+	AttributeKeyInvalidationNonce      = "logic_call_invalidation_nonce"
+	AttributeKeyBadEthSignature        = "bad_eth_signature"
+	AttributeKeyBadEthSignatureSubject = "bad_eth_signature_subject"
```
fedekunze, line 29: rename

module/x/gravity/types/genesis.go
```diff
@@ -67,9 +67,12 @@ var (
 	// ParamsStoreSlashFractionConflictingClaim stores the slash fraction ConflictingClaim
 	ParamsStoreSlashFractionConflictingClaim = []byte("SlashFractionConflictingClaim")
 
-	//  ParamStoreUnbondSlashingValsetsWindow stores unbond slashing valset window
+	// ParamStoreUnbondSlashingValsetsWindow stores unbond slashing valset window
 	ParamStoreUnbondSlashingValsetsWindow = []byte("UnbondSlashingValsetsWindow")
 
+	// ParamStoreSlashFractionBadEthSignature stores the amount by which a validator making a fraudulent eth signature will be slashed
+	ParamStoreSlashFractionBadEthSignature = []byte("SlashFractionBadEthSignature")
```
fedekunze, line 74: ditto

module/x/gravity/types/msgs.go
```diff
@@ -593,3 +594,27 @@ func (msg *MsgCancelSendToEth) GetSigners() []sdk.AccAddress {
 	}
 	return []sdk.AccAddress{sdk.AccAddress(acc)}
 }
+
+// MsgSubmitBadSignatureEvidence
+// ======================================================
+
+// ValidateBasic performs stateless checks
+func (e *MsgSubmitBadSignatureEvidence) ValidateBasic() error {
+	return nil
```
fedekunze, line 603: validate fields

module/x/gravity/types/msgs.go
```diff
@@ -593,3 +594,27 @@ func (msg *MsgCancelSendToEth) GetSigners() []sdk.AccAddress {
...
+}
+
+// GetSignBytes encodes the message for signing
+func (msg MsgSubmitBadSignatureEvidence) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(msg))
```
fedekunze, line 608: no amino please

module/x/gravity/types/msgs.go
```diff
@@ -593,3 +594,27 @@ func (msg *MsgCancelSendToEth) GetSigners() []sdk.AccAddress {
...
+}
+
+// GetSigners defines whose signature is required
+func (msg MsgSubmitBadSignatureEvidence) GetSigners() []sdk.AccAddress {
+	return []sdk.AccAddress{}
+}
+
+// Type should return the action
+func (msg MsgSubmitBadSignatureEvidence) Type() string { return "Submit_Bad_Signature_Evidence" }
```
fedekunze, line 617: no caps

fedekunze (dismissed):
Thanks, @jtremback, I believe this feature needs to be specified first before implementing it. Also, I have some concerns about bloating the state with all the checkpoints and tech debt being introduced with this PR. Happy to guide you through the specification process

jtremback (commented):
Replying to @fedekunze:

> Thanks, @jtremback, I believe this feature needs to be specified first before implementing it. 

This feature has been specified multiple times:
- https://github.com/cosmos/gravity-bridge/blob/main/spec/slashing-spec.md#gravslash-01-signing-fake-validator-set-or-tx-batch-evidence
- https://github.com/althea-net/cosmos-gravity-bridge/issues/207

> Also, I have some concerns about bloating the state with all the checkpoints 

It does bloat state, but not by much. 32 bytes is very tiny in the grand scheme of all the state that gets stored in a blockchain.  Pruning will be easy to add in at a later date. Delaying this feature in order to make micro-optimizations, is IMO not worth it. We discussed this in a standup last week @zmanian

module/proto/gravity/v1/msgs.proto
jtremback: > rename to MsgSubmitMisbehavior

Disagree. "Misbehavior" is very ambiguous. I think there's definitely a better term than "BadSignatureEvidence", but that's not it.

jtremback: > This msg should also contain the signer account address

Ethereum account? Cosmos account?

Neither of these is required. The ethereum address is contained in the signature and is all that is needed to look up the validator. I don't think it's a good practice to bloat out the code with boilerplate that is not actually used by the code.

module/x/gravity/keeper/evidence.go
jtremback: Are you referring to the naming or something else?

module/x/gravity/keeper/evidence.go
jtremback: The logic needs to be run, but only if the subject is one of those 3 types. This pattern was suggested by @jackzampolin on Tuesday

module/x/gravity/keeper/evidence.go
jtremback: I'll just address all of your comments about this here.

I am storing past checkpoints in a set, not an array. From wikipedia: 

"In computer science, a set is an abstract data type that can store unique values, without any particular order. It is a computer implementation of the mathematical concept of a finite set. Unlike most other collection types, rather than retrieving a specific element from a set, one typically tests a value for membership in a set."

It is possible to implement a set using a map, like the KV store provided by Cosmos-SDK. From wikipedia:

"sets are commonly implemented in the same way as (partial) maps (associative arrays) – in this case in which the value of each key-value pair has the unit type or a sentinel value (like 1)"

In this implementation, I am using `[]byte{0x1}` as the sentinel value. Setting the sentinel as the value and the checkpoint as the key allows us to later test whether the checkpoint exists in the set.

module/x/gravity/keeper/evidence.go
jtremback: See my comment about sets below

module/x/gravity/types/batch.go
jtremback: The only way they will panic is through programmer error, which means that it is unnecessary to handle it at runtime. One of the GetCheckpoint functions was already panicking, while the other two were erroring as you see here. When unifying the functions into a single interface, I looked at the code and made a judgement call to use panics. If I am wrong, and there is some way that the errors could happen depending on different conditions at runtime, then we should switch them all to errors.

module/x/gravity/keeper/batch.go
jtremback: See comment below about our use of a set to store checkpoints.

module/x/gravity/keeper/keeper.go
jtremback: Are you referring to the use of the "found" boolean? I pinged you about this in Telegram a few days ago. It was tough to decide whether or not to use the "found" boolean, since it seems that the Cosmos-SDK does not use a consistent pattern. I settled on using "found" since at least it is more explicit.

module/x/gravity/types/msgs.go
jtremback: All of the other functions in the file are using `ModuleCdc.MustMarshalJSON`. It doesn't seem like a good idea to change it for just one function, and refactoring everything would be out of scope for this PR

module/proto/gravity/v1/genesis.proto
jkilpatr: a signature over some batch, validator set, or logic call not already in the store.

module/x/gravity/keeper/evidence.go
fedekunze: I think @jackzampolin suggested to use an interface and remove the switch, which is what makes sense in this case

fedekunze:
> This feature has been specified multiple times

Those aren't specifications, but audit findings and an issue. I'm talking about implementation specifications and architecture decision records. See the SDK [ADRs](https://github.com/cosmos/cosmos-sdk/blob/master/docs/architecture/README.md) and the module [specification](https://github.com/cosmos/cosmos-sdk/blob/master/x/README.md). 

As I said, this PR is **increasing** technical debt in its current state and not specified, which will lead us to more calls to know the rationale of the decisions behind this implementation instead of documenting/spec'ing it from the beginning.

> Delaying this feature in order to make micro-optimizations, is IMO not worth it.

You can ignore my review and merge this PR if you want, but it will not have my approval nor my blessing to be eventually integrated into the Hub

module/proto/gravity/v1/msgs.proto
marbar3778: Why not use MsgSubmitEvidence and use any to describe the type of evidence. This will allow for future iterations of the software to quickly add new types of evidence instead of doing a refactor. 

The design that's being taken could use a design doc, something similar to an ADR.

module/x/gravity/keeper/evidence.go
```diff
@@ -0,0 +1,89 @@
...
+		return k.checkBadSignatureEvidenceInternal(ctx, subject, msg.Signature)
+	case *types.OutgoingLogicCall:
+		return k.checkBadSignatureEvidenceInternal(ctx, subject, msg.Signature)
+
+	default:
+		return sdkerrors.Wrap(types.ErrInvalid, "Bad signature must be over a batch, valset, or logic call")
+	}
+}
+
+func (k Keeper) checkBadSignatureEvidenceInternal(ctx sdk.Context, subject types.EthereumSigned, signature string) error {
```
marbar3778, line 34: nit: checkSignature instead of `checkBadSignatureEvidenceInternal`?

module/x/gravity/keeper/evidence_test.go
jtremback: This file tests:

- The cases where a validator cannot be slashed because the batch, valset, or logic call that they signed actually existed. This tests the code that sets the checkpoint when any of these items is created, as well as the code that looks for one of these checkpoints.
- The case where a validator is slashed successfully for signing a checkpoint that never existed.

It does not test:
- The case where a signature is malformed. I think we should have tests for that elsewhere, instead of testing that in every piece of code that uses our signature code.

marbar3778:
@jtremback do you want to collab on an ADR? I will add the template to this repo shortly

jkilpatr (approved):
While this change does add some technical debt it's a feature that we require and the debt does not outweigh the risks of not having this code.

module/third_party/proto/cosmos/base/query/v1beta1/pagination.proto
```diff
@@ -30,6 +30,9 @@ message PageRequest {
   // count_total is only respected when offset is used. It is ignored when key
   // is set.
   bool count_total = 4;
+
+  // reverse is set to true indicates that, results to be returned in the descending order.
+  bool reverse = 5;
```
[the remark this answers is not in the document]
jkilpatr: probably auto formatting

jtremback:
Dismissing @fedekunze 's review and merging for the following reasons:

- Concerns about no ADR: ADR's have unfortunately not been part of this repo from the beginning. However, I am currently preparing a PR to add ADRs for all significant features and decisions, including this one. Merging allows testing of this feature to move forward in parallel.
- Concerns about state growth from saving checkpoints: The state growth is very small, and is orthogonal to all other security and performance aspects of this feature. Merging this allows us to audit and test all other aspects of this feature. Fixing state growth will be a drop in replacement. 
- Concerns about naming: We are merging this in preparation for doing a codebase-wide rename which implements names specified together with @fedekunze and the Sommelier team. Renaming methods and messages for this feature can take place in context of the broader rename.
- Concerns about "code quality": The PeggyJV branch (which we are contributing to) implements a bunch of nice to have code cleanups. It currently does not include this feature, so merging it into this repo will have no effect on progress there.

PR #356, opened 2021-06-23
malicious supply check - e2e test fixes

levicook:


levicook:
ugh, selected the wrong upstream (again)

PR #364, opened 2021-07-14
word typo

wonderfan:
rectify typo.


blockcreators:
Also, in the main readme:

Gravity is currently can transfer ERC20 assets

Should be:

Gravity can currently transfer ERC20 assets

PR #374, opened 2021-08-23
Jehan/doc cleanup

jtremback:
Cleanup and clarification of some docs


jkilpatr (approved):
Looks good to me. Would recommend maybe putting a link to an index for these spec documents into the documetnation index in the main readme. It took us a little while to find them again, so I'm pretty sure anyone looking for them won't be able to find them easily.

jtremback:
> Looks good to me. Would recommend maybe putting a link to an index for these spec documents into the documetnation index in the main readme. It took us a little while to find them again, so I'm pretty sure anyone looking for them won't be able to find them easily.

I'm planning on doing that in the Althea repo, since that's where most of your docs are

